PR #5, opened 2021-02-25
install-deps.sh Refactoring and Pipeline Improvement

shiyi-peng:
Background: As DO is not planning to publish ARM packages in the near future, we will still need to manually install their dependencies in our pipeline. The main goal of the refactoring is to support different distros, and reduce dependency errors in the future.
* Refactor the install-deps script, utilizing the DO bootstrap script. 
* Add build trigger for main branch
* Fix Bug 32047995: ENABLE_ADU_TELEMETRY_REPORTING- arm64 yaml missing (due to the long time for all three builds, merging this change into this PR)
* The TODO is because DO had a latest change that we need to use, but they haven't tagged a release on this latest commit.
* This PR is only for unblocking the pipeline, the build time improvement will be after this PR


yuvalg-MSFT (dismissed):

scripts/install-deps.sh
```diff
@@ -327,6 +240,12 @@ do_install_do() {
 
     git clone --recursive --single-branch --branch $do_ref --depth 1 $do_url . || return
 
+    if [[ $install_do_deps != "false" ]]; then
+        local bootstrap_file=$do_dir/build/bootstrap/bootstrap-$install_do_deps.sh
+        chmod +x $bootstrap_file || return
```
JeffMill, line 245: they created a install_do_deps script? super cool, if so, but how a dev now use it?  Do they need to install a DO package?  enlist in a DO project?  
Ensure that the OneNote instructions reflect changes.

JeffMill (approved):
:shipit:

scripts/install-deps.sh
shiyi-peng: Yes they created it which is handy for us to use. Do you mean use it in our local dev machine? 
When we install-do with our script, we always clone DO repo, so we will have access to this file. The command will be `--install-do  -install-do-deps ubuntu-18.04`  (or debian-9)
Yes definitely! Will merge the change and do the instructions change in doc.

---
In reply to: [586816521](https://github.com/Azure/iot-hub-device-update/pull/5#discussion_r586816521) [](ancestors = 586816521)

JeffMill (approved):
:shipit:

PR #6, opened 2021-02-26
Added CertificateType to differentiate between x509 and symmetric key

nihemstr:
Added the option for using both the Nested Edge Certificate and x509 support (for EIS)


nihemstr:
Closing instead of rebasing. Significant changes have happened since this version was taken into account.

PR #13, opened 2021-03-11
Task 31853332: Disable ARM 32 Debian9 and Ubuntu ARM 64 builds.

shiyi-peng:
Now that all pipelines are green, we disable ARM 32 Debian9 and Ubuntu ARM 64 builds during daily development.
There is on-going work to improvement the build time of these two builds & refactor the yaml files, whose priority is dependent on the ask from PM of supporting these two arches
* Note: There is a mistake from previous PR on `install-deps.sh`, I added the fix in this PR.

### DIFF
--- a/azurepipelines/adu-debian-arm32-build.yml
+++ b/azurepipelines/adu-debian-arm32-build.yml
@@ -28,7 +28,6 @@ trigger:
   branches:
     include:
       - main
-      # TODO(shiyipeng): Reactivate the pipeline after PM confirm on debian arm support
       # - release/*
   paths:
     exclude:
@@ -45,7 +44,6 @@ pr:
   branches:
     include:
       - main
-      # TODO(shiyipeng): Reactivate the pipeline after PM confirm on debian arm support
       # - release/*
       # - feature/*
       # TODO(shiyipeng): Task 27290511: Debian arm32 64 pipeline - improve build time
@@ -117,7 +115,7 @@ jobs:
           DO_ACCESS_TOKEN: $(do_access_token)
 
       - bash: |
-          sudo docker exec -w /iot-hub-device-update arm32container scripts/install-deps.sh --install-do --install-do-deps debian-9
+          sudo docker exec -w /iot-hub-device-update arm32container scripts/install-deps.sh --install-do --install-do-deps-distro debian-9
         displayName: "Install DO SDK from source in container"
 
         # Cleanup the entries we added to clone private repos.
--- a/azurepipelines/adu-debian-arm32-build.yml
+++ b/azurepipelines/adu-debian-arm32-build.yml
@@ -28,7 +28,8 @@ trigger:
   branches:
     include:
       - main
-      - release/*
+      # TODO(shiyipeng): Reactivate the pipeline after PM confirm on debian arm support
+      # - release/*
   paths:
     exclude:
       - docs/*
@@ -44,8 +45,9 @@ pr:
   branches:
     include:
       - main
-      - release/*
-      - feature/*
+      # TODO(shiyipeng): Reactivate the pipeline after PM confirm on debian arm support
+      # - release/*
+      # - feature/*
       # TODO(shiyipeng): Task 27290511: Debian arm32 64 pipeline - improve build time
       #- dev/*
   paths:
--- a/azurepipelines/adu-ubuntu-arm64-build.yml
+++ b/azurepipelines/adu-ubuntu-arm64-build.yml
@@ -27,7 +27,8 @@ trigger:
   branches:
     include:
       - main
-      - release/*
+      # TODO(shiyipeng): Reactivate the pipeline after PM confirm on ubuntu arm support
+      # - release/*
   paths:
     exclude:
       - docs/*
@@ -43,8 +44,9 @@ pr:
   branches:
     include:
       - main
-      - release/*
-      - feature/*
+      # TODO(shiyipeng): Reactivate the pipeline after PM confirm on ubuntu arm support
+      # - release/*
+      # - feature/*
       # TODO(shiyipeng): Task 27290511: Debian arm32 64 pipeline - improve build time
       #- dev/*
   paths:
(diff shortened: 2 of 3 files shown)


azurepipelines/adu-debian-arm32-build.yml
```diff
@@ -28,7 +28,8 @@ trigger:
   branches:
     include:
       - main
-      - release/*
+      # TODO(shiyipeng): Reactivate the pipeline after PM confirm on debian arm support
+      # - release/*
```
JeffMill, line 32: remove commented lines.  Create task for follow up in future, add note in task as to what to add back into yml.
assign task to PM for now.

azurepipelines/adu-ubuntu-arm64-build.yml
```diff
@@ -27,7 +27,8 @@ trigger:
   branches:
     include:
       - main
-      - release/*
+      # TODO(shiyipeng): Reactivate the pipeline after PM confirm on ubuntu arm support
```
JeffMill, line 30: >TODO [](start = 8, length = 4)

see comment in other yml

JeffMill:
>     install_do_deps=$default_do_deps_distro

should this be "install_do_deps_distro" ?

---
Refers to: scripts/install-deps.sh:383 in 8b19ac4. [](commit_id = 8b19ac46f9e1af7e08b908dc22cb1562477950e9, deletion_comment = False)

JeffMill:
>     if [[ $install_aduc_deps == "true" || $install_do_deps != "false" ]]; then

when/how can this be "false"?  it's a repo/tag name, no?

---
Refers to: scripts/install-deps.sh:403 in 8b19ac4. [](commit_id = 8b19ac46f9e1af7e08b908dc22cb1562477950e9, deletion_comment = False)

JeffMill:
>     install_do_deps=$default_do_deps_distro

it's confusing that we are assigning a distro name to a "boolean"

---
In reply to: [797121568](https://github.com/Azure/iot-hub-device-update/pull/13#issuecomment-797121568) [](ancestors = 797121568)

---
Refers to: scripts/install-deps.sh:383 in 8b19ac4. [](commit_id = 8b19ac46f9e1af7e08b908dc22cb1562477950e9, deletion_comment = False)

JeffMill:
>         if [[ $install_do_deps == "ubuntu-18.04" ]]; then

what happens when we move to ubuntu 20? does this break?

---
Refers to: scripts/install-deps.sh:247 in 8b19ac4. [](commit_id = 8b19ac46f9e1af7e08b908dc22cb1562477950e9, deletion_comment = False)

JeffMill:
>     install_do_deps=$default_do_deps_distro

maybe install_do_deps should be "install_do_deps_distro" and the default is "" (equivalent to false in  the code now).  By doing that, then we have
"": no distro (dont install do deps)
"<name>": distro to use (install do deps)

---
In reply to: [797122167](https://github.com/Azure/iot-hub-device-update/pull/13#issuecomment-797122167) [](ancestors = 797122167,797121568)

---
Refers to: scripts/install-deps.sh:383 in 8b19ac4. [](commit_id = 8b19ac46f9e1af7e08b908dc22cb1562477950e9, deletion_comment = False)

JeffMill:
approved with comments

JeffMill (approved):
:shipit:

shiyi-peng:
>         if [[ $install_do_deps == "ubuntu-18.04" ]]; then

It depends on how DO names their dependency bootstrap file, [https://github.com/microsoft/do-client/tree/main/build/bootstrap](https://github.com/microsoft/do-client/tree/main/build/bootstrap), so this variable needs to sync with the file name. Not sure if this is too fragile - but the reason we do this is that specific to this file we need to add `--no-tools` flag

---
In reply to: [797122345](https://github.com/Azure/iot-hub-device-update/pull/13#issuecomment-797122345) [](ancestors = 797122345)

---
Refers to: scripts/install-deps.sh:247 in 8b19ac4. [](commit_id = 8b19ac46f9e1af7e08b908dc22cb1562477950e9, deletion_comment = False)

shiyi-peng:
>     if [[ $install_aduc_deps == "true" || $install_do_deps != "false" ]]; then

I suspect this might be a little non clear, but it is quite convenient to use. The default being `false` so unless we specify we want install all deps, or install certain DO-deps, we don't install any DO deps. So we don't need two flags like `--install-do-deps --install-do-deps-distro ubuntu-18.04`, instead we can do `--install-do-deps ubuntu-18.04`

---
In reply to: [797121742](https://github.com/Azure/iot-hub-device-update/pull/13#issuecomment-797121742) [](ancestors = 797121742)

---
Refers to: scripts/install-deps.sh:403 in 8b19ac4. [](commit_id = 8b19ac46f9e1af7e08b908dc22cb1562477950e9, deletion_comment = False)

shiyi-peng:
>     install_do_deps=$default_do_deps_distro

I think you are right. this approach is better! I'll make the change

---
In reply to: [797122996](https://github.com/Azure/iot-hub-device-update/pull/13#issuecomment-797122996) [](ancestors = 797122996,797122167,797121568)

---
Refers to: scripts/install-deps.sh:383 in 8b19ac4. [](commit_id = 8b19ac46f9e1af7e08b908dc22cb1562477950e9, deletion_comment = False)

azurepipelines/adu-debian-arm32-build.yml
shiyi-peng: Task 32092462: Confirm on Distro/Arch support breadth, added :smile: 

---
In reply to: [592797799](https://github.com/Azure/iot-hub-device-update/pull/13#discussion_r592797799) [](ancestors = 592797799)

jimson-msft:
Consider also modifying the deb9/ubuntu arm builds to run on a schedule rather than disabling entirely. You'll still receive signals on if they are healthy rather than waiting on re-enabling them.

azurepipelines/adu-debian-arm32-build.yml
```diff
@@ -28,7 +28,7 @@ trigger:
   branches:
     include:
       - main
```
jimson-msft, line 30: This pipeline will still trigger on check-ins to main ( as well as pr's opened against main, which for adu agent is only when dev branches are merged back in correct?) - intended?

PR #16, opened 2021-03-15
Added support for defaulting to the configuration file in the case we…

nihemstr:
- Added support for defaulting to the configuration file in the case we cannot get our connection information from AIS
- Added a printout to show which connection method we're using to connect to the IotHub (Device or Module)

### DIFF
--- a/src/adu_types/src/adu_types.c
+++ b/src/adu_types/src/adu_types.c
@@ -26,7 +26,7 @@ const char* ADUC_ConnType_ToString(const ADUC_ConnType connType)
         return "ADUC_ConnType_Module";
     }
 
-    return "";
+    return "<Unknown>";
 }
 /**
  * @brief DeAllocates the ADUC_ConnectionInfo object
--- a/src/adu_types/src/adu_types.c
+++ b/src/adu_types/src/adu_types.c
@@ -10,6 +10,25 @@
 #include <string.h>
 
 /**
+ * @brief Returns the string associated with @p connType
+ * @param connType ADUC_ConnType to be stringified
+ * @returns if the ADUC_ConnType exists then the string version of the value is returned, "" otherwise
+ */
+const char* ADUC_ConnType_ToString(const ADUC_ConnType connType)
+{
+    switch (connType)
+    {
+    case ADUC_ConnType_NotSet:
+        return "ADUC_ConnType_NotSet";
+    case ADUC_ConnType_Device:
+        return "ADUC_ConnType_Device";
+    case ADUC_ConnType_Module:
+        return "ADUC_ConnType_Module";
+    }
+
+    return "";
+}
+/**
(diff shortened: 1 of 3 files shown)


src/agent/src/main.c
```diff
@@ -724,7 +726,12 @@ _Bool StartupAgent(const ADUC_LaunchArguments* launchArgs)
 #else
         if (!GetConnectionInfoFromIdentityService(&info))
         {
-            goto done;
+            Log_Info("Failed to get connection information from AIS. Defaulting to configuration file");
+
+            if (!GetConnectionInfoFromADUConfigFile(&info))
```
Jerry-GeGarcia, line 731: >GetConnectionInfoFromADUConfigFile [](start = 17, length = 34)

we will now fallback to config? perhaps some opportunity here for some minor refactor later

Jerry-GeGarcia (approved):
:shipit:

src/agent/src/main.c
nihemstr: That's the plan Nox laid out. 

I think we need a clearer story for our provisioning routes. It's becoming more like a web then a straight line.

---
In reply to: [594749103](https://github.com/Azure/iot-hub-device-update/pull/16#discussion_r594749103) [](ancestors = 594749103)

JeffMill: what's the rationale? historically, I've been adverse to having any fallback as it tends to either confuse users, or introduce undesirable behaviors (what if there's a stale config file, and now you're connected to an endpoint you dont want to be connected to?)

---
In reply to: [594754375](https://github.com/Azure/iot-hub-device-update/pull/16#discussion_r594754375) [](ancestors = 594754375,594749103)

src/adu_types/src/adu_types.c
```diff
@@ -9,6 +9,25 @@
...
+    {
+    case ADUC_ConnType_NotSet:
+        return "ADUC_ConnType_NotSet";
+    case ADUC_ConnType_Device:
+        return "ADUC_ConnType_Device";
+    case ADUC_ConnType_Module:
+        return "ADUC_ConnType_Module";
+    }
+
+    return "";
```
JeffMill, line 29: >"" [](start = 11, length = 2)

what do I do elsewhere? Do I send "n/a", or fail (e.g. return null)?  Returning "" can be confusing as if there's a bug it wont be obvious in trace what the value is.

JeffMill (approved):
:shipit:

JeffMill:
approved with comments

src/adu_types/src/adu_types.c
nihemstr: We normally use "<Unknown>"

I'll change this to that.

---
In reply to: [594760708](https://github.com/Azure/iot-hub-device-update/pull/16#discussion_r594760708) [](ancestors = 594760708)

PR #22, opened 2021-03-22
Bug 32212559: Deployment stuck at ‘in progress’: bug in startup code path

shiyi-peng:
When the previous action is Download, and then the device lost connection, when it gets connected and starts up again - it is stuck at `at progress` because the agent reports `idle` in this case, while it should resume the Download workflow.

Add the logic to resume Download workflow as Nox suggested. 

Validation to ensure that this won’t regress anything is needed.

### DIFF
--- a/src/agent/adu_core_interface/src/adu_core_export_helpers.c
+++ b/src/agent/adu_core_interface/src/adu_core_export_helpers.c
@@ -25,7 +25,7 @@
  * @param buffer Where to store identifier.
  * @param buffer_cch Number of characters in @p buffer.
  */
-static void GenerateUniqueId(char* buffer, size_t buffer_cch)
+void GenerateUniqueId(char* buffer, size_t buffer_cch)
 {
     const time_t timer = time(NULL);
     const struct tm* ptm = gmtime(&timer);
--- a/src/agent/adu_core_interface/src/adu_core_export_helpers.c
+++ b/src/agent/adu_core_interface/src/adu_core_export_helpers.c
@@ -8,7 +8,6 @@
 
 #include <stdbool.h>
 #include <stdlib.h>
-#include <time.h>
 
 #include <parson.h>
 
@@ -18,19 +17,6 @@
 #include <aduc/string_c_utils.h>
 #include <sys/wait.h> // for waitpid
 #include <unistd.h>
-
-/**
- * @brief Generate a unique identifier.
- *
- * @param buffer Where to store identifier.
- * @param buffer_cch Number of characters in @p buffer.
- */
-void GenerateUniqueId(char* buffer, size_t buffer_cch)
-{
-    const time_t timer = time(NULL);
-    const struct tm* ptm = gmtime(&timer);
-    (void)strftime(buffer, buffer_cch, "%y%m%d%H%M%S", ptm);
-}
 
 void ADUC_MethodCall_Idle(ADUC_WorkflowData* workflowData);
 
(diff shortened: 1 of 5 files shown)


src/agent/adu_core_interface/src/agent_workflow.c
```diff
@@ -446,6 +446,19 @@ void ADUC_Workflow_HandleStartupWorkflowData(ADUC_WorkflowData* workflowData)
     else
     {
         Log_Info("The installed criteria is not met. The current update was not installed on the device.");
+
+        if (workflowData->CurrentAction == ADUCITF_UpdateAction_Download)
+        {
+            Log_Info("There's a pending 'download' action request. Continue downloading the update.");
+
+            // There's a pending download requrest.
```
nihemstr, line 454: >requrest [](start = 42, length = 8)

Spelling :)

nihemstr (changes requested):
:clock1:

nihemstr:
>     if (IsAducResultCodeFailure(isInstalledResult.ResultCode))

If checking this value covers all possible internal errors then the else{} portion of the logic will not be caught in an infinite loop. Otherwise we'll be stuck trying to download. 

Might need to check that this isn't preventing ADU from reporting potential internall error codes and failures.

---
Refers to: src/agent/adu_core_interface/src/agent_workflow.c:440 in 6ade5c0. [](commit_id = 6ade5c0059972ded19c0a88ff277ea8779b056c2, deletion_comment = False)

src/agent/adu_core_interface/src/agent_workflow.c
```diff
@@ -676,6 +698,11 @@ void ADUC_Workflow_HandleUpdateAction(ADUC_WorkflowData* workflowData)
 
     if (entry->Action == ADUCITF_UpdateAction_Download)
     {
+        // Generate workflowId when we start downloading.
+        GenerateUniqueId(
+            workflowData->WorkflowId, sizeof(workflowData->WorkflowId) / sizeof(workflowData->WorkflowId[0]));
```
Nox-MSFT, line 703: >sizeof(workflowData->WorkflowId) / sizeof(workflowData->WorkflowId[0]) [](http://example.com/codeflow?start=38&length=70)

You could use ARRAY_SIZE macro here..

e.g. 

ARRAY_SIZE(workflowData->WorkflowId)

Nox-MSFT (approved):
:shipit:

src/agent/adu_core_interface/src/adu_core_export_helpers.c
```diff
@@ -774,12 +760,15 @@ void ADUC_MethodCall_Idle(ADUC_WorkflowData* workflowData)
 
         free(workflowData->WorkFolder);
         workflowData->WorkFolder = NULL;
-    }
 
-    // Each time we get to idle, generate a new workflow id.
-    GenerateUniqueId(workflowData->WorkflowId, sizeof(workflowData->WorkflowId) / sizeof(workflowData->WorkflowId[0]));
+        Log_Info("UpdateAction: Idle. Ending workflow with WorkflowId: %s", workflowData->WorkflowId);
+        memset(workflowData->WorkFolder, 0, sizeof(*(workflowData->WorkFolder)));
```
Nox-MSFT, line 765: > memset(workflowData->WorkFolder, 0, sizeof(*(workflowData->WorkFolder))); [](http://example.com/codeflow?start=7&length=74)

Write to NULL address? Segfault?

Nox-MSFT (changes requested):
:clock1:

Nox-MSFT (approved):
:shipit:

shiyi-peng:
>     if (IsAducResultCodeFailure(isInstalledResult.ResultCode))

It is checking (resultCode <= 0); so I think we are covered here?


---
In reply to: [818266782](https://github.com/Azure/iot-hub-device-update/pull/22#issuecomment-818266782) [](ancestors = 818266782)

---
Refers to: src/agent/adu_core_interface/src/agent_workflow.c:440 in 565791b. [](commit_id = 565791b1212d9ea22f1ad20fc6500af8f3fd4629, deletion_comment = False)

nihemstr (approved):
:shipit:

PR #28, opened 2021-03-26
Added generic error handlers to the linux_adu_core_impl

nihemstr:
Added two generic error handlers to the linux_adu_core_impl download step.

Pertains to bug:

https://microsoft.visualstudio.com/OS/_workitems/edit/28240915

### DIFF
--- a/src/platform_layers/linux_platform_layer/src/linux_adu_core_impl.cpp
+++ b/src/platform_layers/linux_platform_layer/src/linux_adu_core_impl.cpp
@@ -186,6 +186,20 @@ ADUC_Result LinuxPlatformLayer::Download(const char* workflowId, const char* upd
 
         extendedResultCode = MAKE_ADUC_DELIVERY_OPTIMIZATION_EXTENDEDRESULTCODE(doErrorCode);
     }
+    catch (const std::exception& e)
+    {
+        Log_Error("DP download failed with an unhandled std exception: %s", e.what());
+
+        resultCode = ADUC_DownloadResult_Failure;
+        extendedResultCode = ADUC_ERC_NOTRECOVERABLE;
+    }
+    catch (...)
+    {
+        Log_Error("DO download failed due to an unknown exception");
+
+        resultCode = ADUC_DownloadResult_Failure;
+        extendedResultCode = ADUC_ERC_NOTRECOVERABLE;
+    }
 
     // If we downloaded successfully, validate the file hash.
     if (resultCode == ADUC_DownloadResult_Success)


src/platform_layers/linux_platform_layer/src/linux_adu_core_impl.cpp
```diff
@@ -186,6 +186,20 @@ ADUC_Result LinuxPlatformLayer::Download(const char* workflowId, const char* upd
 
         extendedResultCode = MAKE_ADUC_DELIVERY_OPTIMIZATION_EXTENDEDRESULTCODE(doErrorCode);
     }
+    catch (const std::exception& e)
+    {
+        Log_Error("DP download failed with an unhandled std exception: %s", e.what());
+
+        resultCode = ADUC_DownloadResult_Failure;
+        extendedResultCode = ADUC_ERC_NOTRECOVERABLE;
```
Nox-MSFT, line 194: Can you check if we should try to get last error (e.g. `errno`) to see if we can return a better error code instead of  ADUC_ERC_NOTRECOVERABLE?

Also, going forward, we should try report 'resultDetails' property w/ any error, if possible.

In this case, e.what() would be very useful for diagnostic.

src/platform_layers/linux_platform_layer/src/linux_adu_core_impl.cpp
```diff
@@ -186,6 +186,20 @@ ADUC_Result LinuxPlatformLayer::Download(const char* workflowId, const char* upd
...
+    }
+    catch (...)
+    {
+        Log_Error("DO download failed due to an unknown exception");
+
+        resultCode = ADUC_DownloadResult_Failure;
+        extendedResultCode = ADUC_ERC_NOTRECOVERABLE;
```
Nox-MSFT, line 201: Let's specify a facility code that can help use distinguish between an error that occurs in 'core' components, vs. error that occurs in 'update content handler'.

Nox-MSFT (changes requested):

src/platform_layers/linux_platform_layer/src/linux_adu_core_impl.cpp
nihemstr: Do we have a resultDetails field? 


---
In reply to: [603721077](https://github.com/Azure/iot-hub-device-update/pull/28#discussion_r603721077) [](ancestors = 603721077)

src/platform_layers/linux_platform_layer/src/linux_adu_core_impl.cpp
nihemstr: This is one of our core components isn't it? The failure doesn't originate from sThe content handlers are going to report their errors in a different way if I remember your spec properly. 

This is coming from a platform_layer. Are we implying that we should have three error sections:

1. AduCoreInterface
2. PlatformLayer
3. ContentHandler

?

That feels like it will be less useful to the user considering their context will be the portal not the actual device.

Does adding the point of failure within our agent add any significant value to the customer when they're vieweing their error log on the portal?

---
In reply to: [603721893](https://github.com/Azure/iot-hub-device-update/pull/28#discussion_r603721893) [](ancestors = 603721893)

Nox-MSFT (approved):

PR #30, opened 2021-03-30
Added ability to read and set the trusted cert option for the Nested …

nihemstr:
Added ability to read and set the trusted cert option for the Nested Edge scenario.

Allows for the scenario of:
1. Reading the certificate from a local path
2. Loading it into the trusted cert option
3. Initiating a connection

### DIFF
--- a/src/adu_types/inc/aduc/adu_types.h
+++ b/src/adu_types/inc/aduc/adu_types.h
@@ -38,7 +38,7 @@ typedef enum tagADUC_AuthType
     ADUC_AuthType_NotSet = 0,
     ADUC_AuthType_SASToken = 1,
     ADUC_AuthType_SASCert = 2,
-
+    ADUC_AuthType_NestedEdgeCert = 3,
 } ADUC_AuthType;
 
 /**
--- a/src/agent/src/main.c
+++ b/src/agent/src/main.c
@@ -637,7 +637,7 @@ _Bool GetConnectionInfoFromADUConfigFile(ADUC_ConnectionInfo* info)
 
     // Optional: The certificate string is needed for Edge Gateway connection.
     if (ReadDelimitedValueFromFile(
-            ADUC_CONF_FILE_PATH, "certificate_path", certificatePath, ARRAY_SIZE(certificatePath)))
+            ADUC_CONF_FILE_PATH, "edgegateway_cert_path", certificatePath, ARRAY_SIZE(certificatePath)))
     {
         if (!LoadBufferWithFileContents(certificatePath, certificateString, ARRAY_SIZE(certificateString)))
         {
--- a/src/agent/src/main.c
+++ b/src/agent/src/main.c
@@ -515,6 +515,15 @@ _Bool ADUC_DeviceClient_Create(ADUC_ConnectionInfo* connInfo, const ADUC_LaunchA
         result = false;
     }
     else if (
+        connInfo->certificateString != NULL && connInfo->authType == ADUC_AuthType_NestedEdgeCert
+        && (iothubResult =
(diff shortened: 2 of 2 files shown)


JeffMill:
>     if (!ClientHandle_CreateFromConnectionString(

this big blob o' else if's should be cleaned up.

are the else ifs here solely to avoid using"goto done" as I do elsewhere, or are they used because there's mutually exclusive options?  line 500-506 seems unrelated to 508+, for example.

---
Refers to: src/agent/src/main.c:493 in 45a384b. [](commit_id = 45a384b8f4795471083c5eb33ddc85e427e30ff3, deletion_comment = False)

JeffMill:
>     if (!ClientHandle_CreateFromConnectionString(

line 508+ seems like it should initially be conditional on authType and then each authType case should check parameters and fail if a required parameter is not specified, rather than falling through to the next case.

---
In reply to: [810616995](https://github.com/Azure/iot-hub-device-update/pull/30#issuecomment-810616995) [](ancestors = 810616995)

---
Refers to: src/agent/src/main.c:493 in 45a384b. [](commit_id = 45a384b8f4795471083c5eb33ddc85e427e30ff3, deletion_comment = False)

JeffMill:
>              g_iotHubClientHandle, ADUC_PnPDeviceTwin_Callback, (void*)g_iotHubClientHandle))

suspicious.  why is a void* cast needed?  

---
Refers to: src/agent/src/main.c:574 in 45a384b. [](commit_id = 45a384b8f4795471083c5eb33ddc85e427e30ff3, deletion_comment = False)

JeffMill (approved):
:shipit:

JeffMill:
approved with comments.

nihemstr:
>     if (!ClientHandle_CreateFromConnectionString(

(some speculation) There's no reason for them. It's just what the IotHub examples use when setting options.

---
In reply to: [810617352](https://github.com/Azure/iot-hub-device-update/pull/30#issuecomment-810617352) [](ancestors = 810617352,810616995)

---
Refers to: src/agent/src/main.c:493 in 45a384b. [](commit_id = 45a384b8f4795471083c5eb33ddc85e427e30ff3, deletion_comment = False)

nihemstr:
>              g_iotHubClientHandle, ADUC_PnPDeviceTwin_Callback, (void*)g_iotHubClientHandle))

the global handle is a userContextCallback value which is the vlaue which will be passed to ADUC_PnPDeviceTwin_Callback.

The standard type for this is void* given that it could technically be whatever you wanted. So the type in the C-SDK is void*. 

You can checkout the actual definition in the client_handle_helper.c for both Module and Device twin callbacks.

---
In reply to: [810617617](https://github.com/Azure/iot-hub-device-update/pull/30#issuecomment-810617617) [](ancestors = 810617617)

---
Refers to: src/agent/src/main.c:574 in 45a384b. [](commit_id = 45a384b8f4795471083c5eb33ddc85e427e30ff3, deletion_comment = False)

nihemstr:
>     if (!ClientHandle_CreateFromConnectionString(

Added a bug to follow up on this:

[https://microsoft.visualstudio.com/OS/_workitems/edit/32306258](https://microsoft.visualstudio.com/OS/_workitems/edit/32306258)

---
In reply to: [810622691](https://github.com/Azure/iot-hub-device-update/pull/30#issuecomment-810622691) [](ancestors = 810622691,810617352,810616995)

---
Refers to: src/agent/src/main.c:493 in 45a384b. [](commit_id = 45a384b8f4795471083c5eb33ddc85e427e30ff3, deletion_comment = False)

PR #40, opened 2021-04-13
Fixing erroneous padding in the arm64 pipeline yaml file

nihemstr:


shiyi-peng (approved):
:shipit:

azurepipelines/adu-ubuntu-arm64-build.yml
```diff
@@ -1,8 +1,8 @@
 variables:
-  version.major: 0
-  version.minor: 7
-  version.patch: 0
-  version.pre-release: "rc1-public-preview"
+    version.major: 0
+    version.minor: 7
```
JeffMill, line 3: >version [](start = 4, length = 7)

yaml is truly horrible. do you know if ADO pipelines support any other format?

JeffMill (approved):
:shipit:

azurepipelines/adu-ubuntu-arm64-build.yml
nihemstr: The official standard is yaml. It's wht all their examples use. If there is another one they do not make it obvious to find.

---
In reply to: [612795016](https://github.com/Azure/iot-hub-device-update/pull/40#discussion_r612795016) [](ancestors = 612795016)

PR #49, opened 2021-04-22
Bug 32763332: LogLevel should include LOG_ERROR in main.c

shiyi-peng:
Expected value for ADUC LOG LEVEL: 0-3
So the >= should be changed to >


JeffMill (approved):
:shipit:

PR #52, opened 2021-04-24
Update how-to-run-agent.md

ValOlson:


docs/agent-reference/how-to-run-agent.md
```diff
@@ -88,3 +88,42 @@ To manually stop the daemon:
 ```shell
 sudo systemctl stop adu-agent
 ```
+
+## How To Create 'adu' Group and User 
+
+Create 'adu' group and 'adu' user by follow these steps:
```
nihemstr, line 94: Is this for people not using the debian package? It would be best to do a callout of that assumption or we break out these instructions into a "how-to-run-a-custom-built-agent.md" Document.

Note: this group needs to exist before installing the AduAgent when using AIS (becasue we need the uid for the adu-agent to create the principal. I ran into a similar issue Saravanan was having on the Santa Cruz agent)

nihemstr (changes requested):
:clock1:

nihemstr (dismissed):
:shipit:

docs/agent-reference/how-to-run-agent.md
nihemstr: Marked as approved consider this a nit.

ValOlson: Please check Nox and I added content in Add agent and Run agent

Nox-MSFT (approved):

nihemstr (approved):
:shipit:

PR #54, opened 2021-04-27
Fix import manifest creator bash script datetime format

devlie:
ADU supposedly accepts ISO 8601 datetime format, but there is a bug in import processing that incorrectly treats offset format as invalid: `2021-04-27T20:56:15+00:00`. Temporarily working around it by using Z designator format: `2021-04-27T20:56:15Z`.

### DIFF
--- a/tools/AduCmdlets/create-adu-import-manifest.sh
+++ b/tools/AduCmdlets/create-adu-import-manifest.sh
@@ -16,7 +16,7 @@ Create an import manifest to import an update into Device Update.
 -v VERSION      Update version
 -t TYPE         Update type
 -i CRITERIA     Installed criteria
--c INFO         Comma separated (manufacturer, model) compatibility information 
+-c INFO         Comma separated (manufacturer, model) compatibility information
                 May be specified multiple times.
 FILE            Path to update file(s)
 "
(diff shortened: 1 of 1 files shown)


tools/AduCmdlets/create-adu-import-manifest.sh
```diff
@@ -221,7 +221,8 @@ for file in "${UPDATE_FILES[@]}"; do
     fi
 done
 
-CREATED_DATETIME=$(date --utc --iso-8601=seconds)
+# ISO 8601 UTC date time with Z designator.
+CREATED_DATETIME=$(date --utc +%FT%TZ)
```
JeffMill, line 225: i'd prefer that we not change this script and we wait until the service is fixed.

changing this script will be confusing as it looks like we DONT support ISO8601, when we will (and should!)

JeffMill (changes requested):
:clock1:

tools/AduCmdlets/create-adu-import-manifest.sh
JeffMill: There is *no* rush to get this working AFAIK.

devlie: This is temporary. We will revert this next week once server side fix is fully deployed. ETA Thursday.

JeffMill (approved):
:shipit:

PR #55, opened 2021-04-28
Bug 32910784: Changing scripts should kick off a pipeline build, 

nihemstr:
- Removed scripts and tools directories from the build pipeline exemption


JeffMill (approved):
:shipit:

PR #61, opened 2021-05-11
Fixed DO group permissions to allow for writing the connection string

nihemstr:
Added the code in our postisnt script to add adu to the do group so we have permissions to write the connection_string to DO in order to fix an issue with the Nested Edge scenario.

### DIFF
--- a/packages/debian/postinst
+++ b/packages/debian/postinst
@@ -152,6 +152,17 @@ case "$1" in
             systemctl restart "$ms_doclient_lite_service"
         fi
 
+        # Note: We must add the 'adu' user to the 'do' group so that we can set the connection_string for
+        # the nested edge scenario. 
+        echo "Add the 'adu' user to the 'do' group"
+        if getent passwd 'adu' > /dev/null; then
+            usermod -aG 'do' 'adu'
+
+            # Restart deliveryoptimization-agent service to ensure that the new 'do' user's permissions take effect.
+            echo "Restart $ms_doclient_lite_service"
+            systemctl restart "$ms_doclient_lite_service"
+        fi
+
         # Do not restart the current adu-agent because we want the update workflow
         # to complete, and reports the agent's state to ADU Management service.
         echo "============================================================"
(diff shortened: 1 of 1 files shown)


Nox-MSFT (changes requested):
:clock1:

packages/debian/postinst
```diff
@@ -152,6 +152,17 @@ case "$1" in
             systemctl restart "$ms_doclient_lite_service"
         fi
 
+        # Note: We must add the 'adu' user to the 'do' group so that we can set the connection_string for
+        # the nested edge scenario. 
+        echo "Add the 'adu' user to the 'do' group"
+        if getent passwd 'adu' > /dev/null; then
```
JeffMill, line 158: we dont own the DO group.  Why are "we" doing this?

JeffMill: related to above where we are adding do user to adu group.

JeffMill (changes requested):
:clock1:

packages/debian/postinst
```diff
@@ -143,9 +143,11 @@ case "$1" in
         # which is one of the dependencies declared in adu-agent control file.
         # We are assuming that both DO user and group currently exist at this point.
         # Add 'do' user to 'adu' group to allow DO to write to ADU download sandbox.
-        echo "Add the 'do' user to the 'adu' group."
+        echo "Add the 'do' user to the 'adu' group and 'adu' to the 'do' group"
         if getent passwd 'do' > /dev/null; then
             usermod -aG 'adu' 'do'
+            # Note: We must add the 'adu' user to the 'do' group so that we can set the connection_string for DO
+            usermod -aG 'do' 'adu'
```
JeffMill, line 150: >adu [](http://example.com/codeflow?start=30&length=3)

`$adu_user` ?

JeffMill (approved):
:shipit:

Nox-MSFT (approved):
:shipit:

PR #70, opened 2021-05-26
Bug 30744239 RemoveInstalledCriteria does not remove all matched entries

jw-msft:
Allow RemoveInstalledCriteria to remove duplicates.

Add tests for removing all dups and not removing non-matching.
Add InstalledCriteriaPersistence RAII class and used it to remove ADUC_INSTALLEDCRITERIA_FILE_PATH on destruction to improve test isolation for tests that use PersistInstalledCriteria().

Update docs for RemoveInstalledCriteria with example json.  

screenshot of doxygen:
image

### DIFF
--- a/src/content_handlers/apt_handler/src/apt_handler.cpp
+++ b/src/content_handlers/apt_handler/src/apt_handler.cpp
@@ -400,7 +400,6 @@ const JSON_Status safe_json_serialize_to_file_pretty(const JSON_Value* value, co
 
 /**
  * @brief Persist specified installedCriteria in a file and mark its state as 'installed'.
- * NOTE: For private preview, only the last installedCriteria is persisted.
  * 
  * @param installedCriteriaFilePath A full path to installed criteria data file.
  * @param installedCriteria An installed criteria string.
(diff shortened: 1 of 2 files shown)


Nox-MSFT (approved):
:shipit:

Nox-MSFT (approved):

src/content_handlers/apt_handler/src/apt_handler.cpp
```diff
@@ -496,7 +505,12 @@ AptHandlerImpl::RemoveInstalledCriteria(const char* installedCriteriaFilePath, c
...
                         success = false;
                     }
-                    break;
+
+                    if (!success)
+                    {
+                        break;
+                    }
+                    // otherwise, keep going to remove duplicates
                 }
```
JeffMill, line 514: do we really want to serialize after every remove (line 500)?  seems like it would be better to move 500-501 outside of the loop.

JeffMill (changes requested):
:clock1:

src/content_handlers/apt_handler/src/apt_handler.cpp
jw-msft: good point!  I will be sure to address this. I am closing this PR and will replicate to align with repo/branching strategy.

PR #74, opened 2021-06-09
PowerShell cmdlet for calling ADU Import Update REST API

devlie:
* Updated `AduUpdate.psm1` module to support Import Manifest v3.
* Added `AduImportUpdate.psm1` module to create import manifest _and_ the import API input. It also uploads update artifacts to AzBlob for staging.
* Added `AduAzStorageBlobHelper.psm1` module to help with uploading files to AzBlob.
* Added `AduRestApi.psm1` module to call ADU Import Update REST API.
* Merged the two READMEs into one so it will appear as soon as user load the directory in GitHub.
* Updated READMEs with new instructions.


dpokluda (changes requested):

JeffMill:
companion bash script?

tools/AduCmdlets/AduAzStorageBlobHelper.psm1
```diff
@@ -0,0 +1,80 @@
...
+        [ValidateNotNullOrEmpty()]
+        $StorageAccountName,
+
+        # Blob container name.
+        [Parameter(Mandatory=$true)]
+        [ValidateNotNullOrEmpty()]
+        $ContainerName,
+
+        # Skip connecting to Azure.
+        [switch] $SkipLogin
```
JeffMill, line 39: what does this do?  Is the assumption that there's a cached connection already?  If so, wouldn't 44-46 be a no-op?

tools/AduCmdlets/AduAzStorageBlobHelper.psm1
```diff
@@ -0,0 +1,80 @@
...
+
+    if (!$SkipLogin)
+    {
+        Connect-AzAccount -ErrorAction Stop -UseDeviceAuthentication | Out-Null
+        $subscription = Get-AzSubscription -SubscriptionId $SubscriptionId -ErrorAction Stop
+        Set-AzContext $subscription -ErrorAction Stop | Out-Null
+    }
+
+    $account = Get-AzStorageAccount -ResourceGroupName $ResourceGroupName -Name $StorageAccountName -ErrorAction Stop
+    $container = New-AzStorageContainer -Name $ContainerName -Context $account.Context -ErrorAction SilentlyContinue
```
JeffMill, line 50: >-ErrorAction SilentlyContinue [](http://example.com/codeflow?start=87&length=29)

why silently continue? I realize that 51 will ultimately fail if this does, but you might mask the root cause of failure, no?

JeffMill:
looks like this uses azure blob storage, so bash might be out (although there is a "cifs" tool).

I think you might be better writing this all in python instead of powershell.  

python is fine cross platform, and azure storage has a [python SDK](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-python)

---
In reply to: [858184608](https://github.com/Azure/iot-hub-device-update/pull/74#issuecomment-858184608) [](http://example.com/codeflow?ancestors=858184608)

tools/AduCmdlets/README.md
```diff
@@ -0,0 +1,209 @@
...
+Get-Help New-AduUpdateCompatibility -Detailed
+Get-Help New-AduImportManifest -Detailed
+Get-Help Get-AduFileHashes -Detailed
+Get-Help New-AduImportUpdateInput -Detailed
+Get-Help Start-AduImportUpdate -Detailed
+Get-Help Wait-AduUpdateOperation -Detailed
+```
+
+## Bash Script
+
```
JeffMill, line 158: i prefer two files.  If I'm using bash why should I wade through half of this document to find what I need?

JeffMill (changes requested):
:clock1:

tools/AduCmdlets/AduAzStorageBlobHelper.psm1
devlie: Skips `connect-azaccount` call. For e.g. to get ref to 2nd container, you can append -SkipLogin. 

`connect-azaccount` doesn't know to cache anything, hence this switch.

tools/AduCmdlets/AduAzStorageBlobHelper.psm1
devlie: new-azstoragecontainer throws error when container already exists.

tools/AduCmdlets/AduImportUpdate.psm1
```diff
@@ -0,0 +1,104 @@
+#
+# Azure Device Update for IoT Hub
+# Powershell module for creating and preparing update for import into Azure Device Update.
```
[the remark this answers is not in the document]
devlie: I think we'll be fine. Our hostname is already **adu**.microsoft.com

tools/AduCmdlets/AduRestApi.psm1
```diff
@@ -0,0 +1,203 @@
...
+    $headers.Add('Content-Type', 'application/json')
+
+    if ($ImportUpdateInput)
+    {
+        $RequestBody = $ImportUpdateInput | ConvertTo-Json -Depth 50
+    }
+
+    Write-Verbose $RequestBody
+
+    $response = Invoke-WebRequest -Uri $url -Method POST -Headers $headers -Body $RequestBody -Verbose:$VerbosePreference
```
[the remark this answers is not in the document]
devlie: I need access to the response headers. Invoke-RestMethod added `-ResponseHeadersVariable` but that's only available from PS6 and script currently requires min PS5.

tools/AduCmdlets/README.md
devlie: Added TOC.

GitHub would render README.md when user navigates to the directory. Having 2 readme files prevent that.

devlie:
Sorry, out of scope for now.

---
In reply to: [858186558](https://github.com/Azure/iot-hub-device-update/pull/74#issuecomment-858186558) [](http://example.com/codeflow?ancestors=858186558,858184608)

PR #85, opened 2021-07-30
Merge release/2021-q2 to main

jw-msft:
Pull commits for 0.8.0rc1 release.


shiyi-peng:
Ok.. DO had some changes on their scripts causing the ARM builds to fail. I can fix them later today

shiyi-peng (approved):

PR #104, opened 2021-12-06
Add ability to install prefix deps

mdavis777:


jw-msft:
FYI. The DO team is working to add support and will integrate this PR in some shape/form once that work is complete.  We also have a work item to do the same in DU Agent--it is already partially done but there is a bit more work to update paths in source code and ensure all installed assets in deb pkg or cmake install are put under the install prefix.

jw-msft:
closing stale pr for now and tracking the task internally.

PR #108, opened 2021-12-10
Split simulator and Delivery Optimization logic

mdavis777:
For users that want a custom interface that doesn't use Delivery
optimization


jw-msft:
Closing stale PR. simulator platform no longer exists and tracking compile-time removal internally.

PR #154, opened 2022-04-29
Fix multi-component update example, improve Steps Handler, Script Handler, and discontinue support for Update Manifest v2

Nox-MSFT:
- Change the installed criteria to match the IsInstalled() implementation in the example script files.
- Clean up Steps Handler implementations and return better resultCode for 'optional' and 'installed' component.
- Update unit tests to match the current Agent implementation. 
- Change Script Handler autogenerated option from **--workfolder** to **--work-folder**.
- Port workflow utils changes to remove support for an update manifest v2. Including relevant unit tests.
- Remove confusing (error) logs when validating existing files in sandbox before download and update affecting code in Hash Utils.


jw-msft (approved):
Signing off for hotfix change made up of custom changes for hotfix to work on hotfix branch. Would prefer to be able to cherry-pick commits to hotfix branch but realize why it needs to be custom retrofitted due to dependent changes and prune out feature work like rollback changes, etc.

PR #157, opened 2022-05-05
Fix memory leak in GetReportingJsonValue() - free root JSON_Value when done

bennydiamond:
Found this bad boy. 
There might still be other memory leaks still lurking around tho. I keep losing memory when I manually retry a failed device update.

### DIFF
--- a/src/agent/adu_core_interface/src/adu_core_interface.c
+++ b/src/agent/adu_core_interface/src/adu_core_interface.c
@@ -820,6 +820,7 @@ _Bool AzureDeviceUpdateCoreInterface_ReportStateAndResultAsync(
     }
 
     jsonString = json_serialize_to_string(rootValue);
+    json_value_free(rootValue);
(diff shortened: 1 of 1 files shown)


src/agent/adu_core_interface/src/adu_core_interface.c
```diff
@@ -820,6 +820,7 @@ _Bool AzureDeviceUpdateCoreInterface_ReportStateAndResultAsync(
     }
 
     jsonString = json_serialize_to_string(rootValue);
+    json_value_free(rootValue);
```
Nox-MSFT, line 823: Thanks for catching this! 
Do you mind move this to line 840 please? In case we need to add some code that uses need rootValue before **done:**

Nox-MSFT (approved):

Nox-MSFT:
Thanks @bennydiamond for your contribution!

PR #169, opened 2022-05-11
Fix false-positive result from script handler when child process failed, and previous result file exists.

Nox-MSFT:
[ADO Bug 39511245](https://microsoft.visualstudio.com/OS/_workitems/edit/39511245): Script Handler return false-positive value when ADUC_LaunchChildProcess exit code != 0, and previous aduc_result file exists

### DIFF
--- a/src/content_handlers/script_handler/src/script_handler.cpp
+++ b/src/content_handlers/script_handler/src/script_handler.cpp
@@ -547,7 +547,6 @@ static ADUC_Result ScriptHandler_PerformAction(const std::string& action, const 
         aduShellArgs.emplace_back(a);
     }
 
-    // Check runtime
     if (IsExtraDebugLogEnabled())
     {
         std::stringstream ss;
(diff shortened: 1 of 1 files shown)


src/content_handlers/script_handler/src/script_handler.cpp
```diff
@@ -559,41 +558,44 @@ static ADUC_Result ScriptHandler_PerformAction(const std::string& action, const
...
+        {
+            result = { .ResultCode = ADUC_Result_Failure,
+                    .ExtendedResultCode = ADUC_ERC_SCRIPT_HANDLER_INSTALL_FAILURE_PARSE_RESULT_FILE };
+            workflow_set_result_details(
+                workflowData->WorkflowHandle,
+                "Cannot parse the script result file '%s'.",
+                scriptResultFile.c_str());
+            goto done;
+        }
+        else
```
jw-msft, line 591: nit: do not need `else` since above did `goto done`

src/content_handlers/script_handler/src/script_handler.cpp
```diff
@@ -559,41 +558,44 @@ static ADUC_Result ScriptHandler_PerformAction(const std::string& action, const
...
     if (exitCode != 0)
     {
         int extendedCode = ADUC_ERC_SCRIPT_HANDLER_CHILD_PROCESS_FAILURE_EXITCODE(exitCode);
         Log_Error(
-            "Child process failed (%s), extendedResultCode:0x%X (exitCode:%d)",
+            "Script failed (%s), extendedResultCode:0x%X (exitCode:%d)",
             action.c_str(),
             extendedCode,
             exitCode);
         result = { .ResultCode = ADUC_Result_Failure, .ExtendedResultCode = extendedCode };
```
jw-msft, line 575: goto done after this line so don't need else on line 577 below either.
Also, should it do workflow_set_result_details like on line 585 too, or no?

src/content_handlers/script_handler/src/script_handler.cpp
```diff
@@ -559,41 +558,44 @@ static ADUC_Result ScriptHandler_PerformAction(const std::string& action, const
...
+                "Cannot parse the script result file '%s'.",
+                scriptResultFile.c_str());
+            goto done;
+        }
+        else
+        {
+            JSON_Object* actionResultObject = json_object(actionResultValue);
+            result.ResultCode = json_object_get_number(actionResultObject, "resultCode");
+            result.ExtendedResultCode = json_object_get_number(actionResultObject, "extendedResultCode");
+            const char* details = json_object_get_string(actionResultObject, "resultDetails");
```
jw-msft, line 596: nit: inline?

src/content_handlers/script_handler/src/script_handler.cpp
```diff
@@ -559,41 +558,44 @@ static ADUC_Result ScriptHandler_PerformAction(const std::string& action, const
...
-        result.ResultCode = json_object_get_number(actionResultObject, "resultCode");
-        result.ExtendedResultCode = json_object_get_number(actionResultObject, "extendedResultCode");
-        const char* details = json_object_get_string(actionResultObject, "resultDetails");
-        workflow_set_result_details(workflowData->WorkflowHandle, details);
+        // Parse result file.
+        actionResultValue = json_parse_file(scriptResultFile.c_str());
+        if (actionResultValue == nullptr)
+        {
+            result = { .ResultCode = ADUC_Result_Failure,
+                    .ExtendedResultCode = ADUC_ERC_SCRIPT_HANDLER_INSTALL_FAILURE_PARSE_RESULT_FILE };
```
jw-msft, line 584: is this the formatting from clang-format ?  Does not seem to line up or be 4 spaces from result =

src/content_handlers/script_handler/src/script_handler.cpp
Nox-MSFT: Fix

src/content_handlers/script_handler/src/script_handler.cpp
Nox-MSFT: Fix. No need for result details here.

src/content_handlers/script_handler/src/script_handler.cpp
Nox-MSFT: Done

src/content_handlers/script_handler/src/script_handler.cpp
Nox-MSFT: Fixed

jw-msft (approved):

PR #170, opened 2022-05-11
Fix download regression where download api returned 'success' when file has is invalid

Nox-MSFT:
**Issue 1** 
ResultCode was set to 'success' when the file validation failed after download. This is a regression from previous code refactoring/cleanup.

**Issue 2**
When preparing steps' workflow for an update that contains multiple steps. We didn't reset childHandle at the begging of the loop. That handler will be freed if any error occurs, but not removed from the parent handler. This causes double-freed of the same handler later on, when we free the parent handler.


src/extensions/content-downloaders/deliveryoptimization-downloader/deliveryoptimization-content-downloader.cpp
```diff
@@ -152,7 +152,8 @@ ADUC_Result do_download(
             fullFilePath.str().c_str(),
             ADUC_HashUtils_GetHashValue(entity->Hash, entity->HashCount, 0),
             algVersion,
-            true);
+            false);
```
jw-msft, line 155: add ` /* suppressErrorLog */ ) inline comment after arg value

src/extensions/extension_manager/src/extension_manager.cpp
```diff
@@ -719,13 +717,13 @@ ADUC_Result ExtensionManager::Download(
     {
         // If target file exists, validate file hash.
         // If file is valid, then skip the download.
-        isFileExistsAndValidHash = ADUC_HashUtils_IsValidFileHash(
+        bool validHash = ADUC_HashUtils_IsValidFileHash(
             childManifestFile.str().c_str(),
             ADUC_HashUtils_GetHashValue(entity->Hash, entity->HashCount, 0),
             algVersion,
             false);
```
jw-msft, line 724: here too

src/extensions/extension_manager/src/extension_manager.cpp
```diff
@@ -740,36 +738,31 @@ ADUC_Result ExtensionManager::Download(
...
-        {
-            result = downloadProc(entity, workflowId, workFolder, retryTimeout, downloadProgressCallback);
-        }
-        catch (...)
+        result = downloadProc(entity, workflowId, workFolder, retryTimeout, downloadProgressCallback);
+    }
+    catch (...)
+    {
+        result = { .ResultCode = ADUC_Result_Failure,
+                    .ExtendedResultCode = ADUC_ERC_CONTENT_DOWNLOADER_DOWNLOAD_EXCEPTION };
```
jw-msft, line 748: Indent doesn't seem to line up here. Run clang-format ?

src/extensions/extension_manager/src/extension_manager.cpp
```diff
@@ -740,36 +738,31 @@ ADUC_Result ExtensionManager::Download(
...
+        goto done;
+    }
+
+    if (IsAducResultCodeSuccess(result.ResultCode))
+    {
+        if (!ADUC_HashUtils_IsValidFileHash(
+                childManifestFile.str().c_str(),
+                ADUC_HashUtils_GetHashValue(entity->Hash, entity->HashCount, 0),
+                algVersion,
+                true))
```
jw-msft, line 758: add ` /* suppressErrorLog */ ) inline comment after arg value here too

src/extensions/extension_manager/src/extension_manager.cpp
```diff
@@ -740,36 +738,31 @@ ADUC_Result ExtensionManager::Download(
...
+        result = { .ResultCode = ADUC_Result_Failure,
+                    .ExtendedResultCode = ADUC_ERC_CONTENT_DOWNLOADER_DOWNLOAD_EXCEPTION };
+        goto done;
+    }
+
+    if (IsAducResultCodeSuccess(result.ResultCode))
+    {
+        if (!ADUC_HashUtils_IsValidFileHash(
+                childManifestFile.str().c_str(),
+                ADUC_HashUtils_GetHashValue(entity->Hash, entity->HashCount, 0),
```
jw-msft, line 756: Since GetHashValue can return NULL in case entity Hash/HashCount was corrupt for some reason (buffer overrun, etc), would be good to call it before and set result + goto done rather than passing NULL into IsValidFilehash.

Also, add `/* index */` after the zero value last arg.

src/utils/hash_utils/src/hash_utils.c
```diff
@@ -264,7 +264,7 @@ _Bool ADUC_HashUtils_IsValidFileHash(const char* path, const char* hashBase64, S
         };
     }
 
-    success = GetResultAndCompareHashes(&context, hashBase64, algorithm, !suppressErrorLog, NULL);
+    success = GetResultAndCompareHashes(&context, hashBase64, algorithm, suppressErrorLog, NULL);
```
jw-msft, line 267: add `'/* outputHash */` after the NULL

src/extensions/content-downloaders/deliveryoptimization-downloader/deliveryoptimization-content-downloader.cpp
Nox-MSFT: done

src/utils/hash_utils/src/hash_utils.c
Nox-MSFT: done

src/extensions/extension_manager/src/extension_manager.cpp
Nox-MSFT: done

src/extensions/extension_manager/src/extension_manager.cpp
Nox-MSFT: Fixed

jw-msft (approved):

jw-msft (approved):

PR #183, opened 2022-06-08
[0.8.x] compilation: add option for hardening flags, defaults to ON.

neonsoftware:
When compiling for an embedded system, such as OpenWRT,
the hardening settings are global to the whole system.

In such case one needs to disable the specific flags that are currently
forced in the current build of this application, in order to avoid conflicts.

The new switch is ON by default, therefore this path is retro-compatible.

Since this patch, though, one can set the option to OFF in the OpenWRT/Yocto recipe
in order to let the hardening settings common to the whole system to be applied.


neonsoftware:
@jw-msft This PR has also been rebased on develop and reformatted as #184

PR #184, opened 2022-06-09
[0.8.x] adu-shell: remove hardcoded paths, use ADUSHELL_FILE_PATH instead.

neonsoftware:
The macro ADUSHELL_FILE_PATH is already properly setup and provided by the build system.
Although hardcoded paths were still in source files.

Since the user might customise its value, use such the ADUSHELL_FILE_PATH instead of hardcoded
values all along the project.

### DIFF
--- a/src/content_handlers/CMakeLists.txt
+++ b/src/content_handlers/CMakeLists.txt
@@ -38,6 +38,8 @@ add_subdirectory (swupdate_handler)
 #   add_subdirectory (my_handler)
 #
 
+target_compile_definitions (${PROJECT_NAME} PUBLIC ADUSHELL_FILE_PATH="${ADUSHELL_FILE_PATH}")
+
 if (${ADUC_PLATFORM_LAYER} STREQUAL "simulator")
     target_compile_definitions (${PROJECT_NAME} PUBLIC ADUC_SIMULATOR_MODE=1)
 endif ()
--- a/src/content_handlers/script_handler/CMakeLists.txt
+++ b/src/content_handlers/script_handler/CMakeLists.txt
@@ -31,7 +31,8 @@ target_link_libraries (
             aduc::system_utils
             aduc::workflow_data_utils
             aduc::workflow_utils
-            -zdefs
-            )
+            -zdefs)
+
+target_compile_definitions (${target_name} PUBLIC ADUSHELL_FILE_PATH="${ADUSHELL_FILE_PATH}")
 
 install (TARGETS ${target_name} LIBRARY DESTINATION ${ADUC_EXTENSIONS_INSTALL_FOLDER})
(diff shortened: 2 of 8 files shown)


src/content_handlers/CMakeLists.txt
```diff
@@ -38,6 +38,8 @@ add_subdirectory (swupdate_handler)
 #   add_subdirectory (my_handler)
 #
 
+target_compile_definitions (${PROJECT_NAME} PUBLIC ADUSHELL_FILE_PATH="${ADUSHELL_FILE_PATH}")
```
jw-msft, line 41: apt_handler.cpp also uses `adushconst::adu_shell` and `#include`'s `adushell_const.hpp`, so I would expect this `target_compile_definitions` to be in `src/content_handlers/apt_handler/CMakeLists.txt` also.

jw-msft:
Thanks for the contribution. Just a couple of asks:
- Please target `develop` branch with the PR rather than `main`. We cherry-pick contributions from develop into private repo and it will appear in next minor/major release rev to public `main`. Main branch requires full suite of testing(currently internal). 
- If haven't done already, please run ./scripts/cmake-format.sh and ./scripts/clang-format.sh when changes are in working set or staged so it might need something like `git reset --soft @~1` before running.
 
Note: packages/debian/postinst also has hard-coded path to adu-shell but that was already the case and ok as the .deb pkg is for binary core agent (currently also includes 1st party extensions).

src/content_handlers/CMakeLists.txt
neonsoftware: Done.

neonsoftware:
>     * Please target `develop` branch with the PR rather than `main`. We cherry-pick contributions from develop into private repo and it will appear in next minor/major release rev to public `main`. Main branch requires full suite of testing(currently internal).

Done.

>     * If haven't done already, please run ./scripts/cmake-format.sh and ./scripts/clang-format.sh when changes are in working set or staged so it might need something like `git reset --soft @~1` before running.

No, I didn't already. Done.

neonsoftware:
@jw-msft Just pinging here to know if the fixes were satisfactory

src/content_handlers/script_handler/CMakeLists.txt
```diff
@@ -31,7 +31,8 @@ target_link_libraries (
             aduc::system_utils
             aduc::workflow_data_utils
             aduc::workflow_utils
-            -zdefs
-            )
+            -zdefs)
+
+target_compile_definitions (${target_name} PUBLIC ADUSHELL_FILE_PATH="${ADUSHELL_FILE_PATH}")
```
jw-msft, line 36: nit: Was there are reason it is PUBLIC? I think PRIVATE should be sufficient and is more consistent with others.

PR #190, opened 2022-06-16
[0.8.x] swupdate: fix execution logic in apply action.

neonsoftware:
Previously the value ADUC_Result_Apply_RequiredImmediateReboot was *always* set,
even in case of error.

This means that:

1) In case of bad exit code, the flow was:

   if (exitCode != 0)
   {
       ...
       result = { ADUC_Result_Failure, exitCode };
       goto done;

    done:
       result = { ADUC_Result_Apply_RequiredImmediateReboot };
       return result;

The issue here is that the first error result assignement is *overwritten* by the second seuccess_reboot defautl result.

2) In case of canceled request:

    // Cancel requested?
    if (workflow_get_operation_cancel_requested(workflowData->WorkflowHandle))
    {
        CancelApply(ADUC_LOG_FOLDER); // Other options for result could be ADUC_Result_Success or ADUC_Result_InProgress
        goto done;

    done:
       result = { ADUC_Result_Apply_RequiredImmediateReboot };
       goto done;

    The issue here is that:
    a - A reboot is requested (successful result) even if the operation was canceled.
    b - The CancelApply result is never returned to the workflow.
    c - DOUBT: I have assumed that returning the CancelApply result to the workflow is the inital expected behavior in the design,
        but it might actually even be that it should be returned to the workflow ADUC_Result_Success or ADUC_Result_InProgress and the
        result of the Cancel operation be unused. In such case the cancel might be handled by the workflow instead.

### DIFF
--- a/src/content_handlers/swupdate_handler/src/swupdate_handler.cpp
+++ b/src/content_handlers/swupdate_handler/src/swupdate_handler.cpp
@@ -200,9 +200,9 @@ ADUC_Result SWUpdateHandlerImpl::Install(const tagADUC_WorkflowData* workflowDat
         args.emplace_back(data.str().c_str());
 
         args.emplace_back(adushconst::target_log_folder_opt);
-        
+
         // Note: this implementation of SWUpdate Handler is relying on ADUC_LOG_FOLDER value from build pipeline.
-        // For GA, we should make this configurable in du-config.json. 
+        // For GA, we should make this configurable in du-config.json.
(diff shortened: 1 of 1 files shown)


neonsoftware:
Hi, 

I had found an unexpected behavior in the apply flow, although, I am not flow will match the architecture. I will let you provide your opinion on whether the current flow was the expected one ot which way to do. Therefore creating it as draft.

I have **added details on the issue in the commit description**. 

Have a nice day.

PR #239, opened 2022-11-03
Fix issue in reboot/restart flow and race condition in D2C messaging util.

Nox-MSFT:
- Fix an issue for image-based update scenario where the device doesn't reboot after the new image is installed.
- Add ADUC_D2C_Message_Type_Device_Properties to avoid a race condition.

### DIFF
--- a/src/utils/d2c_messaging/src/d2c_messaging.c
+++ b/src/utils/d2c_messaging/src/d2c_messaging.c
@@ -526,12 +526,12 @@ bool ADUC_D2C_Message_SendAsync(
     {
         if (s_pendingMessageStore[type].completedCallback != NULL)
         {
-            Log_Debug("Replacing existing pending message. (t:%d)", type);
+            Log_Debug("Replacing existing pending message. (t:%d, s:%s)", type, s_pendingMessageStore[type].content);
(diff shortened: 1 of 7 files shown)


src/utils/d2c_messaging/src/d2c_messaging.c
```diff
@@ -526,12 +526,12 @@ bool ADUC_D2C_Message_SendAsync(
...
         if (s_pendingMessageStore[type].completedCallback != NULL)
         {
-            Log_Debug("Replacing existing pending message. (t:%d)", type);
+            Log_Debug("Replacing existing pending message. (t:%d, s:%s)", type, s_pendingMessageStore[type].content);
             OnMessageProcessingCompleted(&s_pendingMessageStore[type], ADUC_D2C_Message_Status_Replaced);
         }
     }
 
-    Log_Debug("Queueing message (t:%d, c:0x%x", type, message);
+    Log_Debug("Queueing message (t:%d, c:0x%x, t:%s)", type, message, message);
```
jw-msft, line 534: format for 3rd arg is `t:%s`
Did you mean to have value for 3rd arg be essentially `tostring(type)`, or should it be `s:%s` for the type's pending msg store content?

PR #249, opened 2022-11-11
Revert clang-format.sh and disable from precommit git hook

jw-msft:
Revert clang-format.sh due to issues with dry-run of staged content on some versions and disable in precommit githook for now due to incompatible sets of cmd-line argument across versions.

### DIFF
--- a/scripts/clang-format.sh
+++ b/scripts/clang-format.sh
@@ -30,27 +30,11 @@ if [ -z "$GITROOT" ]; then
     $ret 1
 fi
 
-issue_with_file=false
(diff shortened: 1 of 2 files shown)


nihemstr (approved):
Approved. 

When are we looking at having this run again? It would be good to be formatting regularly during our commits

jw-msft:
> Approved.
> 
> When are we looking at having this run again? It would be good to be formatting regularly during our commits

Shooting for some time this week, but looking into better way of doing rather than homegrown hook script.

PR #250, opened 2022-11-15
Fix install-deps issue when --work-folder used on ubuntu 20.04

jw-msft:
Repro'ed on fresh ubuntu 20.04 "quick create" and 
`. ./scripts/install-deps.sh -k -f ~/adu-deps -a`

Fix by:

- Explicitly create and set owner/permissions on the work folder
- Remove unnecessary $SUDO calls
- Make shellcheck (and cmake) symlinks be under /tmp instead of ${work_folder} s.t.
  - No need to pass --cmake-path to build.sh
  - and so precommit git hook script will continue to work when calling sh-format.sh -v

### DIFF
--- a/scripts/install-deps.sh
+++ b/scripts/install-deps.sh
@@ -34,7 +34,8 @@ install_packages=false
 install_packages_only=false
 # The folder where source code will be placed
 # for building and installing from source.
-work_folder=/tmp
+DEFAULT_WORKFOLDER=/tmp
+work_folder=$DEFAULT_WORKFOLDER
 keep_source_code=false
 use_ssh=false
 
@@ -62,7 +63,7 @@ install_cmake_version="$supported_cmake_version"
 cmake_force_source=false
 cmake_prefix="$work_folder"
 cmake_installer_dir=""
-cmake_dir_symlink="${work_folder}/deviceupdate-cmake"
+cmake_dir_symlink="/tmp/deviceupdate-cmake"
 
 install_shellcheck=false
 supported_shellcheck_version='0.8.0'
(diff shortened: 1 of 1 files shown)


nihemstr (approved):
Approved

scripts/install-deps.sh
```diff
@@ -852,6 +854,13 @@ while [[ $1 != "" ]]; do
     shift
 done
 
+# setup workfolder if different from default
```
JeffMill, line 857: im not sure i understand why this is needed. shouldn't a mkdir provide appropriate rights?

scripts/install-deps.sh
```diff
@@ -666,13 +668,13 @@ do_install_shellcheck() {
             return $ret_val
         fi
 
-        ln -sf "${HOME}/.cabal/bin/shellcheck" "${work_folder}/deviceupdate-shellcheck" || return 1
+        ln -sf "${HOME}/.cabal/bin/shellcheck" "/tmp/deviceupdate-shellcheck" || return 1
```
JeffMill, line 671: previously we had an alias for the target folder.  now you're hardcoding /tmp here and on 677

scripts/install-deps.sh
```diff
@@ -617,15 +618,16 @@ do_install_cmake_from_installer() {
     fi
 
     local fullpath_cmake_installer_sh="${work_folder}/${cmake_installer_sh}"
-    $SUDO chmod u+x "${fullpath_cmake_installer_sh}"
-    $SUDO "${fullpath_cmake_installer_sh}" --include-subdir --skip-license --prefix=${cmake_prefix}
+    $SUDO chown "$(id -un)":"$(id -gn)" "${fullpath_cmake_installer_sh}"
```
JeffMill, line 621: not sure i understand why chown is needed -- can you explain?

scripts/install-deps.sh
```diff
@@ -62,7 +63,7 @@ install_cmake_version="$supported_cmake_version"
 cmake_force_source=false
 cmake_prefix="$work_folder"
 cmake_installer_dir=""
-cmake_dir_symlink="${work_folder}/deviceupdate-cmake"
+cmake_dir_symlink="/tmp/deviceupdate-cmake"
```
JeffMill, line 66: see comment below -- you're now hardcoding  /tmp

JeffMill (changes requested):

Nox-MSFT (approved):

PR #236, opened 2022-11-02
fix where updateIdStr gets created

danewalton:
makes same change which was created here: https://github.com/Azure/iot-hub-device-update/pull/215/files

PR #253, opened 2022-11-17
Removing PowerShell cmdlets (replace with CLI)

andrewbrownmsft:
As of GA, the Azure CLI has powerful import-related functionality that goes beyond just creating import manifests. The CLI is also more commonly used cross-platform relative to PowerShell, and the CLI is actively supported with new features to be added in the future. We therefore want to direct folks to use the CLI instead of the existing PowerShell scripts for any import tasks, and are removing the scripts to avoid any confusion.

[Note: this is my first attempt at a PR in this repo, so if something isn't as expected, a detailed explanation of what exactly the problem is and how to address it will be welcomed.]


devlie:
Not sure what Azure:develop branch is for, but is this really safe to do? There seems to be at least a few customers who are using these based on PRs and issues filed. There's at least one who's using it in their E2E test (see https://github.com/Azure/iot-hub-device-update/pull/236)

devlie (changes requested):
Not sure what Azure:develop branch is for, but is this really safe to do? There seems to be at least a few customers who are using these based on PRs and issues filed. There's at least one who's using it in their E2E test (see https://github.com/Azure/iot-hub-device-update/pull/236)

andrewbrownmsft:
> Not sure what Azure:develop branch is for, but is this really safe to do? There seems to be at least a few customers who are using these based on PRs and issues filed. There's at least one who's using it in their E2E test (see #236)

This was intended to target the develop branch as described here: https://microsoft.sharepoint.com/teams/DSPlat/_layouts/15/Doc.aspx?sourcedoc={0e47909a-0b93-4fab-8946-864c0395b352}&action=edit&wd=target%28GitHub.one%7C8e550a3c-b423-4f9d-b70f-a96bdb61a1ee%2FHowTo%20Integrate%20External%20Contributor%20changes%20into%20develop%7Cdcbfc005-29a9-4a19-96e9-910922bffb0d%2F%29&wdorigin=703

If we do have active usage, I am fine with keeping them, and will close this in favor of modifying the README.md file to promote the CLI as our recommended solution.

PR #265, opened 2022-11-29
sort paths before asserting in systemutils_forEachDir tests

jw-msft:


src/utils/system_utils/tests/system_utils_ut.cpp
```diff
@@ -330,9 +329,13 @@ TEST_CASE_METHOD(TestCaseFixture, "SystemUtils_ForEachDir")
...
-        std::stringstream ss;
-        ss << "42, " << TestPath() << ", " << test_subdir << std::endl;
-        std::string expected = ss.str();
+        std::vector<std::string> expected;
+
+        {
+            std::stringstream ss;
+            ss << "42, " << TestPath() << ", " << test_subdir << std::endl;
+            expected.push_back(ss.str());
+        }
```
JeffMill, line 338: im confused. you have a vector of expected paths, but you're still using a comma separated stringsteram here?

src/utils/system_utils/tests/system_utils_ut.cpp
```diff
@@ -341,31 +344,30 @@ TEST_CASE_METHOD(TestCaseFixture, "SystemUtils_ForEachDir")
...
+        };
+
+        verify_dirs_fn();
 
         int err = SystemUtils_ForEachDir(
             TestPath() /* baseDir */, nullptr /* excludeDir */, &functor /* perDirActionFunctor */);
         REQUIRE(err == 0);
 
-        CHECK_THAT(actual, Equals(expected));
+        std::sort(actual.begin(), actual.end(), std::less<std::string>());
```
JeffMill, line 361: perhaps add a wrapper in this UT, `vector<string> SystemUtils_ForEachDir_Sorted()` which has a local "vector<string>"  calls SystemUtils_ForEachDir and then calls sort on the vector and returns it.  Would be a nice C++ wrapper for that method and would likely simplify some of the additions you've made here.

JeffMill: as if your expected vector is already sorted, you should be able to use a single line:

`CHECK(ForEachDir_Sorted(.., ..) == vector({ "a", "b", "c" });`

JeffMill (changes requested):

src/utils/system_utils/tests/system_utils_ut.cpp
jw-msft: This internal, test string format in perDirFn callback, is just capturing the 3 arguments being passed to it by SystemUtils_ForEach.

Each time the perDirFn is called, it will push a vector of the 3 args concatenated in this format.

jw-msft: I've created a ForEachDirCallRecord struct to explicitly model the callback call capture instead of string concatenation.

src/utils/system_utils/tests/system_utils_ut.cpp
jw-msft: I've implemented it. The SystemUtils_ForEachDir takes a functor that has C-style callback function and context member as function object state.

Therefore, I created an static AddRecord method in the TestCaseFixture where it passes the context (which is the TestCaseFixture object instance) and the ForEachDirCallRecord object. 

The actual TEST_CASE_METHOD SECTION test code is now just 2-3 lines each.  Thanks for your suggestions!

src/utils/system_utils/tests/system_utils_ut.cpp
```diff
@@ -9,81 +9,214 @@
...
+ * @param second The second call record object.
 
-static void perDirFn(void* context, const char* baseDir, const char* subdir_name)
+ * @return true when equal.
+ */
+bool operator==(const ForEachDirCallRecord& first, const ForEachDirCallRecord& second)
 {
-    callback_called = true;
-    std::stringstream ss;
+    return first.context == second.context && first.baseDir == second.baseDir && first.subDir == second.subDir;
```
JeffMill, line 37: comparing context pointers? wouldn't you want to compare the context objects?

src/utils/system_utils/tests/system_utils_ut.cpp
```diff
@@ -9,81 +9,214 @@
...
 
-static bool callback_called = false;
-std::string actual;
+/**
+ * @brief A capture of the arguments passed in a call invocation to callback for each sub dir in base dir.
+ *
+ */
+struct ForEachDirCallRecord
+{
+    void* context;
```
JeffMill, line 22: see my comment below -- this might be better as a concrete object (or typed object pointer) rather than void* as you *know* what type this will be in this test.  if you're trying to do something reusable (which is good), maybe make this templated, but that latter is overkill for a UT

JeffMill (approved):
approved with suggestions

src/utils/system_utils/tests/system_utils_ut.cpp
jw-msft: created bug to capture these 2 remaining comments.

src/utils/system_utils/tests/system_utils_ut.cpp
jw-msft: Normally yes, but for the purposes of the tests it just wants to verify that the obj ptr of the recorded call matches what was expected.

src/utils/system_utils/tests/system_utils_ut.cpp
jw-msft: Fixed in PR: https://github.com/Azure/iot-hub-device-update/pull/308

PR #266, opened 2022-11-30
Unbuntu Core - configure hook, extensions registration, and folders owner/user changes

Nox-MSFT:
Changes
- Use SNAPCRAFT_* environment variable in snapcraft.yaml
- Update document
- Remove install and post-refresh hooks
- Update configure hook to create required directories and register DU extensions
- Create extension registration dirs using current user id and group id instead of 'adu'
- Soft-linking DU folders to appropriate SNAP_* folders
- **Curl Content Downloader**
	- Add a build option to support `snap` environment (executing curl from $SNAP/usr/bin/curl)
- **Script Handler**
	- Add an option to run script using effective user (bypassing adu-shell)


jw-msft (approved):

docs/agent-reference/how-to-build-deviceupdate-agent-snap.md
```diff
@@ -63,23 +63,100 @@ tree squashfs-root
...
+```shell
+$ sudo snap run --shell deviceupdate-agent
+```
+
+Optionally, copy the prepared `du-config.json` from [host machine] home directory to `$SNAP_DATA/config` directory
 
-```sh
- snap run --shell deviceupdate-agent
+```shell
+$ cp <home-dir>/du-config.json $SNAP_DATA/config
```
shiyi-peng, line 154: I used `vi`, but a little unsure about the `cp` here. Is it run inside the device-update snap shell? how do we specify the <home-dir>?

Nox-MSFT: Yes, this should be run inside the snap shell. Customer can choose their preferred method to author the du-config.json. As long as it is placed in the right location ($SNAP_DATA/config/du-config.json)

shiyi-peng (approved):
Approved with one comment

PR #274, opened 2022-12-04
Added parsing of the protected properties into a json string for hash…

nihemstr:
- Added function for parsing the protected properties from the json into a string so we can hash it for signature validation
- Added uts for string parsing

### DIFF
--- a/src/utils/rootkeypackage_utils/src/rootkeypackage_parse.c
+++ b/src/utils/rootkeypackage_utils/src/rootkeypackage_parse.c
@@ -830,7 +830,7 @@ ADUC_Result RootKeyPackage_ParseProtectedPropertiesString(JSON_Object* rootObj, 
         goto done;
     }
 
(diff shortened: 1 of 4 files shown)


src/utils/rootkeypackage_utils/src/rootkeypackage_parse.c
```diff
@@ -818,6 +818,54 @@ ADUC_Result RootKeyPackage_ParseProtectedProperties(JSON_Object* rootObj, ADUC_R
...
+        goto done;
+    }
+
+    const char* protectedPropertiesString = json_serialize_to_string(protectedPropertiesValue);
+
+    protectedPropertiesStringHandle = STRING_construct(protectedPropertiesString);
+
+    if (protectedPropertiesStringHandle == NULL)
+    {
+        goto done;
```
jw-msft, line 847: set result.ExtendedResultCode to ADUC_ERC_NOMEM ?

jw-msft (approved):

PR #282, opened 2022-12-08
Remove curl dependency in https_proxy_utils (Bug 42504722)

JeffMill:


src/utils/https_proxy_utils/src/https_proxy_utils.c
```diff
@@ -53,109 +121,85 @@ bool InitializeProxyOptions(HTTP_PROXY_OPTIONS* proxyOptions)
...
+        *end = '\0';
 
-    if (port != NULL)
-    {
-        curl_free(port);
+        proxyOptions->port = atoi(end + 1);
     }
 
-    if (host != NULL)
+    proxyOptions->host_address = strdupe(start);
```
jw-msft, line 192: If we do not want to preclude customers who use international domain names who have built libcurl with IDN support, then need to do convert percent-decoded that correspond to UTF-8 bytes sequences (1 - 3 characters) as per [RFC3986](https://www.rfc-editor.org/rfc/rfc3986).

For example "%E3%82%A2" should become the KATAKANA LETTER A
and "%C3%80" should become LATIN CAPITAL LETTER A WITH GRAVE.

on cmdline, curl built with IDN support can download this: `curl http://www.malmö.se/file`

Here is Curl_idnconvert_hostname function to do IDN conversion of hostname:
https://github.com/curl/curl/blob/eb559c8056916f1e71570ac058d599347bebc2b4/lib/url.c#L1574

## Excerpt from RFC3986
```
"The reg-name syntax allows percent-encoded octets in order to
   represent non-ASCII registered names in a uniform way that is
   independent of the underlying name resolution technology.  Non-ASCII
   characters must first be encoded according to UTF-8 [[STD63](https://www.rfc-editor.org/rfc/rfc3986#ref-STD63)], and then
   each octet of the corresponding UTF-8 sequence must be percent-
   encoded to be represented as URI characters.  URI producing
   applications must not use percent-encoding in host unless it is used
   to represent a UTF-8 character sequence."
```

## libcurl can handle international domain names
https://curl.se/libcurl/c/CURLOPT_URL.html has this to say:

### Encoding
```
"The string pointed to in the [CURLOPT_URL](https://curl.se/libcurl/c/CURLOPT_URL.html) argument 
is generally expected to be a sequence of characters using an ASCII compatible encoding.

If libcurl is built with IDN support, the server name part of the URL can use an "international name" by 
using the current encoding (according to locale) or UTF-8 (when winidn is used; or a Windows Unicode 
build using libidn2).

If libcurl is built without IDN support, the server name is used exactly as specified when passed to the
 name resolver functions."
```

src/utils/https_proxy_utils/src/https_proxy_utils.c
```diff
@@ -9,12 +9,84 @@
 #include "aduc/logging.h"
 #include "aduc/string_c_utils.h"
 #include <azure_c_shared_utility/crt_abstractions.h> // mallocAndStrcpy_s
```
jw-msft, line 11: this doesn't seem to be used anymore

src/utils/https_proxy_utils/src/https_proxy_utils.c
```diff
@@ -9,12 +9,84 @@
...
 #include <strings.h>
 
+/**
+ * @brief Implementation of the non-standard strdup
+ * @param proxyOptions str string to duplicate
+ * @return char* duplicated string, deallocate using free. NULL on failure.
+ */
+static char* strdupe(const char* str)
+{
+    char* dupe = malloc(strlen(str) + 1);
```
jw-msft, line 24: strlen will crash if str is NULL. From usage here, it looks like that'll never happen, but still...

jw-msft (approved):

PR #292, opened 2022-12-14
Removed log message for retrieving connection information from AIS 

nihemstr:
Removed the logging prompt from the GetConnectionInfoFromEIS() function because it can cause confusion when called by other sections than  main.c. The caller should be the one to inform the user not the function being called.

Related bug: https://microsoft.visualstudio.com/OS/_queries/edit/41718105/?triage=true

### DIFF
--- a/src/agent/src/main.c
+++ b/src/agent/src/main.c
@@ -1030,8 +1030,6 @@ bool GetConnectionInfoFromIdentityService(ADUC_ConnectionInfo* info)
     }
     memset(info, 0, sizeof(*info));
 
-    Log_Info("Requesting connection string from the Edge Identity Service");
-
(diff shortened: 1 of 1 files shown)


src/agent/src/main.c
```diff
@@ -1030,8 +1030,6 @@ bool GetConnectionInfoFromIdentityService(ADUC_ConnectionInfo* info)
     }
     memset(info, 0, sizeof(*info));
 
-    Log_Info("Requesting connection string from the Edge Identity Service");
-
```
Nox-MSFT, line 1034: Note: Feel free to move **GetConnectionInfoFrom*** into **communication_abtraction** or [**create new**] communication_util/manager

Nox-MSFT (approved):

src/agent/src/main.c
nihemstr: That's what I'm thinking of doing. Separate bug though

PR #296, opened 2022-12-15
Bug 42164115: Script handler unconditionally downloads primary script payload zero for IsInstalled eval and sometimes script is deleted failing deployment

shiyi-peng:


yuvalg-MSFT (approved):

src/extensions/step_handlers/script_handler/src/script_handler.cpp
```diff
@@ -134,25 +139,76 @@ static ADUC_Result Script_Handler_DownloadPrimaryScriptFile(ADUC_WorkflowHandle
...
+        result.ExtendedResultCode = ADUC_ERC_CONTENT_DOWNLOADER_FILE_HASH_TYPE_NOT_SUPPORTED;
         goto done;
     }
 
-    try
+    // If file exists and has a valid hash, then skip download.
+    // Otherwise, delete an existing file, then download.
+    Log_Debug("Check whether '%s' has already been download into the work folder.", scriptFilePath.c_str());
+
+    if (access(scriptFilePath.c_str(), F_OK) == 0)
```
jw-msft, line 161: As Nox pointed out offline, seems redundant with same check here: https://github.com/Azure/iot-hub-device-update/blob/main/src/extensions/extension_manager/src/extension_manager.cpp#L883

I think the root issue may be something different that is causing script file to be deleted.

src/extensions/step_handlers/script_handler/src/script_handler.cpp
```diff
@@ -134,25 +139,76 @@ static ADUC_Result Script_Handler_DownloadPrimaryScriptFile(ADUC_WorkflowHandle
...
 
-        result = ExtensionManager::Download(entity, handle, &downloadOptions, nullptr);
+        fileExist = true;
     }
-    catch (...)
+
+    if (!fileExist)
     {
-        result.ExtendedResultCode = ADUC_ERC_SCRIPT_HANDLER_DOWNLOAD_PRIMARY_FILE_FAILURE_UNKNOWNEXCEPTION;
+        createResult = ADUC_SystemUtils_MkSandboxDirRecursive(workFolder);
```
jw-msft, line 192: I feel that work folder should be created by extension manager download; otherwise, every content handler has to dup logic to create it before calling download.

jw-msft (changes requested):

jw-msft:
I determined that It's a race condition and sharing violation with the script that is executed by adu-shell for, say, IsInstalled and then when it goes to verify file hash of script again before calling again for Download, it gets this:

[pid 682540] 17:05:54 openat(AT_FDCWD, "/var/lib/adu/downloads/7a5296a2-df0f-4361-9e9c-d825c7076295/swupdatev2_handler_script.sh", O_RDONLY) = -1 EACCES (Permission denied)
...
...
...
2023-01-11T01:05:54.8562Z 682540[682540] [E] Cannot open file: /var/lib/adu/downloads/7a5296a2-df0f-4361-9e9c-d825c7076295/swupdatev2_handler_script.sh, errno: 13 [ADUC_HashUtils_IsValidFileHash]

jw-msft:
Closing this one for now as it does not fix the issue.

PR #297, opened 2022-12-15
One line fix on how-to-build-agent-code.md

shiyi-peng:
Currently, the agent cannot build without having DO as a dependency. To avoid failures, I think we should make the modification on the documentation

### DIFF
--- a/docs/agent-reference/how-to-build-agent-code.md
+++ b/docs/agent-reference/how-to-build-agent-code.md
@@ -49,7 +49,7 @@ To install all dependencies run:
 To install only the dependencies necessary for the agent:
 
 ```shell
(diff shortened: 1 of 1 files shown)


docs/agent-reference/how-to-build-agent-code.md
```diff
@@ -49,7 +49,7 @@ To install all dependencies run:
 To install only the dependencies necessary for the agent:
 
 ```shell
-./scripts/install-deps.sh --install-aduc-deps --install-packages
+./scripts/install-deps.sh --install-aduc-deps --install-packages --install-do
```
nihemstr, line 52: Looks good to me!

nihemstr (approved):

Issue #212, opened 2022-09-13
adu-agent v0.8.2 could not handle iothub reprovisioning


JoergZeidler:
Hi,
after a device re-provisioning process, the adu-agent no longer connects to the iothub.
My setup looks like this:
- adu-agent: v0.8.2
- iotedge 1.4

The log shows the process:
At 8:50:06 the iot-identity and the adu-agent starting.
At 8:50:09 the adu-agent started successfully and the module twin connection state is set to "connected".
At 9:07:25 reprovisioning is triggered using (iotedge system reprovision).
From 9:07:35 adu-agent can no longer connect to the iothub. 
Since the adu-agent does not handle this state, the connection to the iot-hub remains marked as "disconnected" from this point in time.

Here is my Log from adu-agent and aziot-identityd
````
[root@raspberrypi4-64 ics-dm]# journalctl -u aziot-identityd -u adu-agent -b0 --no-pager
Sep 13 08:50:06 raspberrypi4-64 systemd[1]: Started Azure IoT Identity Service.
Sep 13 08:50:06 raspberrypi4-64 systemd[1]: Starting Device Update Agent daemon....
Sep 13 08:50:06 raspberrypi4-64 aziot-identityd[911]: 2022-09-13T08:50:06Z [INFO] - Starting service...
Sep 13 08:50:06 raspberrypi4-64 aziot-identityd[911]: 2022-09-13T08:50:06Z [INFO] - Version - dev build
Sep 13 08:50:06 raspberrypi4-64 aziot-identityd[911]: 2022-09-13T08:50:06Z [INFO] - Provisioning starting. Reason: Startup
Sep 13 08:50:06 raspberrypi4-64 aziot-identityd[911]: 2022-09-13T08:50:06Z [INFO] - Provisioned with backup for xx:xx:xx:xx:xx:xx.
Sep 13 08:50:06 raspberrypi4-64 aziot-identityd[911]: 2022-09-13T08:50:06Z [INFO] - Successfully provisioned with DPS.
Sep 13 08:50:06 raspberrypi4-64 aziot-identityd[911]: 2022-09-13T08:50:06Z [INFO] - Provisioning complete.
Sep 13 08:50:06 raspberrypi4-64 aziot-identityd[911]: 2022-09-13T08:50:06Z [INFO] - Identity reconciliation started. Reason: Startup
[...]

PR #298, opened 2022-12-16
Try to reconnect to IoT Hub on bad credential or expired SAS token (when using AIS)

Nox-MSFT:
- Try to reconnect to IoT Hub on bad credential or expired SAS token (when using AIS)
- Fix static analysis errors in main.c
- Move all iothub connection related codes into a new iothub_connection_manager lib
- Improve iothub connection status handler logics
- Some minor code cleanup

Issue addressed: #212


src/utils/d2c_messaging/inc/aduc/d2c_messaging.h
```diff
@@ -23,43 +23,47 @@ EXTERN_C_BEGIN
...
-    ADUC_D2C_Message_Status_Pending = 0,            /**< Waiting to be processed */
-    ADUC_D2C_Message_Status_In_Progress,            /**< Being processed by the messages processor */
-    ADUC_D2C_Message_Status_Waiting_For_Response,   /**< Sent to the cloud, waiting for response */
-    ADUC_D2C_Message_Status_Success,                /**< Message has been processed successfully */
-    ADUC_D2C_Message_Status_Failed,                 /**< A failure occurred. No longer process */
-    ADUC_D2C_Message_Status_Replaced,               /**< Message was replaced by a new one */
-    ADUC_D2C_Message_Status_Canceled,               /**< Message was canceled */
-    ADUC_D2C_Message_Status_Max_Retries_Reached,    /**< Maximum number of retries reached */
+    ADUC_D2C_Message_Status_Pending = 0, /**< Waiting to be processed */
+    ADUC_D2C_Message_Status_In_Progress, /**< Being processed by the messages processor */
```
nihemstr, line 37: nit: Explicit settings can make it easier to debug some enums (eg the connection reasons) if these are ever made into umock enums for unit testing

src/communication_managers/iothub_communication_manager/src/iothub_communication_manager.c
```diff
@@ -0,0 +1,714 @@
...
+    ADUC_ConfigInfo_UnInit(&config);
+    return succeeded;
+}
+
+/**
+ * @brief Get the Connection Info from Identity Service
+ *
+ * @return true if connection info can be obtained
+ */
+static bool GetConnectionInfoFromIdentityService(ADUC_ConnectionInfo* info)
```
nihemstr, line 519: nit: For all these functions but IotHub_CommunicationManager_<name> so it's easier to reference

src/communication_managers/iothub_communication_manager/src/iothub_communication_manager.c
```diff
@@ -0,0 +1,714 @@
...
+
+        g_next_authentication_attempt_time = (nextRetryTime);
+        Log_Info("The connection is currently broken. Will try to authenticate in %d seconds.", nextRetryTime - now_time);
+        return;
+    }
+
+    // Try to authenticate.
+    g_last_authentication_attempt_time = now_time;
+    g_authentication_retries++;
+    ADUC_Refresh_IotHub_Connection_SAS_Token();
```
nihemstr, line 700: This isn't productive for connections that are made manually. Did I miss a check for the type of connection we're trying to make? 

In the end it shouldn't matter but if we're seeing a "sas token expired" on a connection that is using shared access keys (eg manual connection strings) we should err out or make that log. That's an unexpected case that should be handled by the C-SDK and not the agent at large.  

That could almost be a bug or a log line instead of a larger feature change.

nihemstr (approved):
Awesome addition. Approved.

src/agent/src/main.c
```diff
@@ -691,13 +691,36 @@ static void ADUC_ConnectionStatus_Callback(
     {
         switch (reason)
         {
+        default:
+            break;
         case IOTHUB_CLIENT_CONNECTION_EXPIRED_SAS_TOKEN:
-            Log_Warn("IotHub connection SAS token expired. Attempting refresh.");
+        case IOTHUB_CLIENT_CONNECTION_BAD_CREDENTIAL:
```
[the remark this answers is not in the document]
Nox-MSFT: This code was removed.

src/communication_managers/iothub_communication_manager/src/iothub_communication_manager.c
Nox-MSFT: As discussed, for the 'string' connection type, when the authentication broken, we should try to re-connect, in case that the root cause is fixed by the solution operator.

src/agent/src/main.c
```diff
@@ -691,13 +691,36 @@ static void ADUC_ConnectionStatus_Callback(
...
+        {
+            Log_Warn("IotHub connection broken (reason:%d), refreshing the connection info.");
+            ADUC_ConfigInfo config = {};
+            if (!ADUC_ConfigInfo_Init(&config, ADUC_CONF_FILE_PATH))
+            {
+                Log_Error("Cannot read configuration file.");
```
yuvalg-MSFT: I think we should also log any results if we have it

src/agent/src/main.c
```diff
@@ -888,10 +905,10 @@ static bool ADUC_DeviceClient_Create(
         result = true;
     }
 
-    if ((result == false) && (g_iotHubClientHandle != NULL))
+    if ((result == false) && (*clientHandle != NULL))
```
yuvalg-MSFT: should check the pointer before dereferencing it?

yuvalg-MSFT (approved):

nihemstr (approved):

PR #310, opened 2023-01-04
fix zlog file handle leak

jw-msft:
Ran valgrind with file descriptor traching:
sudo -u adu $VALGRIND_BIN --track-fds=all /usr/bin/AducIotAgent -l0 -e

Before the fix it leaks the agent zlog init'ed from agent/main.c
After the fix, valgrind shows that it does not leak it.

### DIFF
--- a/src/agent/src/main.c
+++ b/src/agent/src/main.c
@@ -1185,7 +1185,6 @@ int main(int argc, char** argv)
     // high-privileged tasks, such as, registering agent's extension(s).
     if (!RunAsDesiredUser())
     {
-        ret = 1;
         goto done;
     }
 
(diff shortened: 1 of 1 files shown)


src/agent/src/main.c
```diff
@@ -1203,7 +1224,8 @@ int main(int argc, char** argv)
         {
             Log_Error("Agent health check failed.");
         }
-        return healthy ? 0 : 1;
+        ret = healthy ? 0 : 1;
```
yuvalg-MSFT, line 1227: I think that if you initialize ret as 0, then you don't have to set it to zero everytime you goto done above. You only set it to 1 before the done tag

yuvalg-MSFT (dismissed):

yuvalg-MSFT (approved):

PR #315, opened 2023-01-05
Fix deadlock in zlog flush deinit

jw-msft:


nihemstr (approved):

src/logging/zlog/src/zlog.c
```diff
@@ -48,16 +48,12 @@ static const char level_names[] = { 'D', 'I', 'W', 'E' }; // Must align with ZLO
 static FILE* zlog_fout = NULL;
 static char* zlog_file_log_dir = NULL;
 static char* zlog_file_log_prefix = NULL;
+static time_t zlog_last_flushed;
```
JeffMill, line 51: init this.

JeffMill: looks like zlog_init does, but doesn't hurt to set to 0.

src/logging/zlog/src/zlog.c
```diff
@@ -375,81 +368,14 @@ void zlog_log(enum ZLOG_SEVERITY msg_level, const char* func, const char* fmt, .
...
-{
-    zlog_buffer_flush_thread_run = true;
-    if (pthread_create(&_zlog_flush_thread, NULL, zlog_buffer_flush_thread, NULL) == 0)
+    if (msg_level == ZLOG_ERROR || (seconds - zlog_last_flushed) >= ZLOG_FLUSH_INTERVAL_SEC)
     {
-        _is_flush_thread_initialized  = true;
-    }
-}
+        _zlog_buffer_lock();
+        _zlog_flush_buffer();
```
JeffMill, line 374: zlog_flush_buffer should do the lock and unlock, no?

JeffMill (changes requested):

src/logging/zlog/src/zlog.c
jw-msft: line 353 where it rolls the log seems to want to hold the lock for longer so I think that's why _zlog_flush_buffer() has a comment that says to acquire the lock before calling.

JeffMill (approved):

PR #319, opened 2023-01-06
Move to popen, pclose. Process utils ut working.

JeffMill:


src/utils/process_utils/src/process_utils.cpp
```diff
@@ -43,115 +44,47 @@
...
-        childExitStatus = WTERMSIG(wstatus);
-        Log_Info("Child process terminated, signal %d", childExitStatus);
-    }
-    else if (WCOREDUMP(wstatus))
+    while (fgets(buffer, sizeof(buffer), fp) != nullptr)
     {
-        // Child process produced a core dump
-        childExitStatus = WCOREDUMP(wstatus);
-        Log_Error("Child process terminated, core dump %d", childExitStatus);
+        output += buffer;
```
jw-msft, line 78: need to handle calling ADUC_pclose on exception thrown.  RAII wrapper would be ideal instead of goto done.

src/utils/process_utils/src/process_utils.cpp
```diff
@@ -43,115 +44,47 @@
...
     }
-    else
+    // Returns 0 if no error occurred.
+    ret = ferror(fp);
+    if (ret != 0)
     {
-        childExitStatus = EXIT_FAILURE;
-        // Child process terminated abnormally.
-        Log_Error("Child process terminated abnormally.", childExitStatus);
+        return ret;
```
jw-msft, line 84: need to handle calling ADUC_pclose to cleanup when failed.

src/utils/process_utils/src/process_utils.cpp
```diff
@@ -43,115 +44,47 @@
...
-        argv.emplace_back(nullptr);
-
-        // The exec() functions only return if an error has occurred.
-        // The return value is -1, and errno is set to indicate the error.
-        int status = ADUCPAL_execvp(command.c_str(), &argv[0]);
-
-        fprintf(stderr, "execvp failed, returned %d, error %d\n", status, errno);
-
-        _exit(EXIT_FAILURE);
+        redirected_command += arg;
```
jw-msft, line 61: not sure but would using std stringstream for redirected_command potentially reduce heap reallocations?

src/libaducpal/src/stdio.c
```diff
@@ -4,27 +4,26 @@
...
+#    include <stdio.h> // popen, pclose
 
 #    include <errno.h> // ENOTDIR, _get_errno
 
 FILE* ADUCPAL_popen(const char* command, const char* type)
 {
-    __debugbreak();
-    return NULL;
+    // TODO(JeffMill): [PAL] Will this return utf8? If not, additional work is needed.
+    return _popen(command, type);
```
jw-msft, line 14: why is there underscore on front?  Also why not pclose(stream) instead of _pclose(stream) on line 19?

src/utils/process_utils/src/process_utils.cpp
```diff
@@ -21,6 +19,9 @@
 #include <aduc/config_utils.h>
 #include <aduc/logging.h>
 #include <aduc/string_utils.hpp>
+
+#include <aducpal/stdio.h> // popen,pclose
```
jw-msft, line 23: super nit: needs ADUCPAL_ prefixes or omit

src/utils/process_utils/src/process_utils.cpp
JeffMill: possibly.  in addition, a prealloc of the string buffer might be an option.  I'll check

src/utils/process_utils/src/process_utils.cpp
JeffMill: fixed

src/utils/process_utils/src/process_utils.cpp
JeffMill: we didn't before -- if we're out of memory, we should probably just crash, no?  otherwise I can wrap this with a try catch.  what do you think?

src/libaducpal/src/stdio.c
JeffMill: underscore is the POSIX compatible name.

src/utils/process_utils/src/process_utils.cpp
JeffMill: my comments indicate the underlying api.

src/utils/process_utils/src/process_utils.cpp
jw-msft: probably not worth it as an exception here indicates oom

PR #321, opened 2023-01-07
Implement nftw.  All UTs passing.

JeffMill:


src/libaducpal/src/ftw.c
```diff
@@ -165,318 +78,124 @@ static int construct_typeflag(const DWORD attrib, int flags)
...
+
+    const int typeflag = typeflag_from_stat(&st, flags);
+
+    // If func() returns nonzero, then the tree walk is terminated and the value returned by fn()
+    // is returned as the result of nftw().
+    return func(fpath, &st, typeflag, ftwbuf);
+}
+
+// Return 0 on success, -1 if an error occurs.
+static int do_nftw(const char* path, NFTW_FUNC_T func, int nopenfd, int flags, int level)
```
jw-msft, line 100: I would expect the impl to start closing older FDs when it gets past some threshold so that it doesn't keep too many FDs open.

src/libaducpal/src/ftw.c
```diff
@@ -165,318 +78,124 @@ static int construct_typeflag(const DWORD attrib, int flags)
...
-				 */
-                (void)fd_limit;
-                lev.level = new.level;
-                lev.base = (int)wcslen(dir->path);
-
-                const wchar_t* thisname = dir->ffd.cFileName;
-
-                /* skip directories named '.' and '..' */
-                if (thisname[0] == L'.' && (!thisname[1] || (thisname[1] == L'.' && !thisname[2])))
+                ret = do_nftw(subdir, func, nopenfd, flags, level + 1);
```
jw-msft, line 138: recursive is not ideal. a non-recursive approach that queues subdirs would be better approach to keep the callstack depth shallower.

src/libaducpal/src/ftw.c
jw-msft: I may have missed it, but I don't see any logic ensuring nopenfd as a max number of FDs to have opened--therefore, the process could run out of FDs.  It's supposed to start closing older FDs that age out of the FD buffer with `nopenfd` number of elements as it recurses down the directory tree.

As per the manpage https://linux.die.net/man/3/nftw:
> To avoid using up all of the calling process's file descriptors, nopenfd specifies the maximum number of directories that ftw() will hold open simultaneously. When the search depth exceeds this, ftw() will become slower because directories have to be closed and reopened. ftw() uses at most one file descriptor for each level in the directory tree.

src/libaducpal/src/ftw.c
```diff
@@ -165,318 +78,124 @@ static int construct_typeflag(const DWORD attrib, int flags)
...
-    wcscpy_s(pathbuf, sizeof(pathbuf) / sizeof(pathbuf[0]), wpath);
-
-    make_forward_slash(pathbuf);
-
-    return do_wnftw(pathbuf, func, fd_limit, flags, NULL);
+    return ret;
 }
 
+// Returns 0 on success, and -1 if an error occurs.
 int ADUCPAL_nftw(const char* path, NFTW_FUNC_T func, int nopenfd, int flags)
```
jw-msft, line 197: there's findFilesInDir in file_utils that uses DIRENT, so it could be tweaked/ported to C to allow returning "DIR" file types as well as regular files and a C wrapper on top of the cpp module could be added to adapt it.

Or better yet, you could break away from nftw semantics and create a "foreach_file" API that just calls the passed-in callback after stat'ing each file that would include all types of files -- directory, regular files, symlinks, fifo pipes, etc.

src/libaducpal/src/ftw.c
JeffMill: yeah, i'll add a TODO for respecting nopenfd - I meant to put that in there.  I don't think it's critical for us right now.

src/libaducpal/src/ftw.c
JeffMill: Perhaps -- do you know how large (in K) a typical Linux callstack is?  I've used this about 8 levels deep and didn't run into any issues.

src/libaducpal/src/ftw.c
JeffMill: findFilesInDir doesn't handle subdirs, so that doesn't add any additional value here.  opendir() is the "standard" way of doing this, which is what im using

PR #327, opened 2023-01-12
Fix overwrite of contract maj ver in downloaders and comp enumerator

jw-msft:


JeffMill (approved):

src/extensions/extension_manager/src/extension_manager.cpp
```diff
@@ -602,8 +602,7 @@ void ExtensionManager::_FreeComponentsDataString(char* componentsJson)
         goto done;
     }
 
-    if (ExtensionManager::_componentEnumeratorContractVersion.majorVer == ADUC_V1_CONTRACT_MAJOR_VER
```
JeffMill, line 605: Did ADUC_ContractUtils_IsV1Contract exist already?

JeffMill (approved):

src/extensions/extension_manager/src/extension_manager.cpp
jw-msft: Yes, it did exist. Should've been using it

PR #328, opened 2023-01-12
Remove command helper in WIOT

JeffMill:
Not needed in WIOT.

### DIFF
--- a/src/agent/CMakeLists.txt
+++ b/src/agent/CMakeLists.txt
@@ -8,7 +8,11 @@ add_subdirectory (adu_core_export_helpers)
 add_subdirectory (adu_core_interface)
 add_subdirectory (device_info_interface)
 add_subdirectory (pnp_helper)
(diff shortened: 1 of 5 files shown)


src/agent/CMakeLists.txt
```diff
@@ -8,7 +8,11 @@ add_subdirectory (adu_core_export_helpers)
 add_subdirectory (adu_core_interface)
 add_subdirectory (device_info_interface)
 add_subdirectory (pnp_helper)
-add_subdirectory (command_helper)
+
+# TODO(JeffMill): [PAL] I don't think Windows needs command_helper.  Check with Jeff or Nox.
+if (NOT WIN32)
```
jw-msft, line 13: It should probably stay aligned at least longer term as this API will be used for future scenarios.  mkfifo is just a named pipe so should have an equivalent in win32 but the linux quirkiness of named pipe may lead to differences in windows for the message boundaries.

PR #330, opened 2023-01-12
Added input params checks to ConnectionStringUtils_IsNestedEdge to prevent test failures

nihemstr:
- Added an input param check for ConnectionStringUtils_IsNestedEdge to prevent unit test error messages
- Bug for PR: https://microsoft.visualstudio.com/OS/_queries/edit/42934385/?triage=true

### DIFF
--- a/src/utils/c_utils/src/connection_string_utils.c
+++ b/src/utils/c_utils/src/connection_string_utils.c
@@ -48,6 +48,11 @@ bool ConnectionStringUtils_DoesKeyExist(const char* connectionString, const char
 {
     bool keyExists = false;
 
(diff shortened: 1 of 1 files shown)


src/utils/c_utils/src/connection_string_utils.c
```diff
@@ -48,6 +48,11 @@ bool ConnectionStringUtils_DoesKeyExist(const char* connectionString, const char
 {
     bool keyExists = false;
 
+    if (connectionString == NULL || key == NULL)
```
jw-msft, line 51: IsNullOrEmpty?

Nox-MSFT (approved):

PR #333, opened 2023-01-14
Fix bug in swupdate-handler v2 example

Nox-MSFT:
- Fix bug in the SWUpdate Handler V2 example scripts
- Fix some typos in SWUpdate Handler V2 and Script Hanlder

### DIFF
--- a/src/extensions/step_handlers/swupdate_handler_v2/tests/testdata/adu-yocto-ab-rootfs-update/payloads/example-a-b-update.sh
+++ b/src/extensions/step_handlers/swupdate_handler_v2/tests/testdata/adu-yocto-ab-rootfs-update/payloads/example-a-b-update.sh
@@ -52,9 +52,9 @@ log_error_pref="\033[1;31m[E]\033[0m"
 # Files and Folders information
 #
 workfolder=
-swu_file=
 output_file=/adu/logs/swpupdate.output
 log_file=/adu/logs/swupdate.log
+swupdate_log_file=/adu/logs/swupdate_log_file
 result_file=/adu/logs/swupdate.result.json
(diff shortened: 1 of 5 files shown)


src/extensions/step_handlers/swupdate_handler_v2/tests/testdata/adu-yocto-ab-rootfs-update/payloads/example-a-b-update.sh
```diff
@@ -222,9 +222,10 @@ print_help() {
     echo "==========================="
     echo ""
     echo "--workfolder              A work-folder (or sandbox folder)."
-    echo "--swu-file                A swu file to install."
+    echo "--swu-file, --image-file  An image file (.swu) file to install."
```
jw-msft, line 225: makes it seem like you can use --swu-file OR --image-file but I am not seeing the `--image-file)` below. I only see  `--swu-file)` but it uses `$image_file` instead of `$swu_file`

src/extensions/step_handlers/swupdate_handler_v2/tests/testdata/adu-yocto-ab-rootfs-update/payloads/example-a-b-update.sh
```diff
@@ -52,9 +52,9 @@ log_error_pref="\033[1;31m[E]\033[0m"
 # Files and Folders information
 #
 workfolder=
-swu_file=
```
jw-msft, line 55: should we put `image_file=` here?

yuvalg-MSFT (approved):

PR #343, opened 2023-01-18
Ensure that the step's extendedResultCode is valid when failed to load its desired handler

Nox-MSFT:


yuvalg-MSFT (approved):
this looks rather straightforward but Jeff Miller should probably approve this as well. Also clearly this needs to be tested to see whether it fixes Jeff's original issue

JeffMill (approved):

PR #350, opened 2023-01-19
fix readdir EOVERFLOW for qemu arm32 on 64bit host

jw-msft:
If inode does not fit into 32-bit offset, then getdirents in glic will return errno 75 EOVERFLOW
It only repros on 32-bit arm running in qemu/docker on top of 64-bit host os.
Compiling with -D_FILE_OFFSET_BITS=64 seems to fix the issue.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -9,6 +9,11 @@ if (ADUC_WARNINGS_AS_ERRORS)
     set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Werror")
 endif ()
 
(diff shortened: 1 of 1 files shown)


nihemstr (approved):

yuvalg-MSFT (approved):

jw-msft:
From https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html
Macro: _FILE_OFFSET_BITS
...
On 64 bit systems this macro has no effect since the *64 functions are identical to the normal functions.
...

PR #358, opened 2023-01-24
Incrementing Pipeline Version to match agent version

nihemstr:
- Incrementing the version to 1.0.1 to reflect the new patch, doesn't effect agent functionality just allows our ARM64 pipelines to output the right version

Bug for PR: https://microsoft.visualstudio.com/OS/_queries/edit/43033501/?triage=true


src/agent/adu_core_interface/src/adu_core_interface.c
```diff
@@ -120,7 +120,7 @@ void ADUC_WorkflowData_Uninit(ADUC_WorkflowData* workflowData)
 /**
  * @brief Reports the client json via PnP so it ends up in the reported section of the twin.
  *
- * @param messageType The message type.
+ * @param messageType The message type.`
```
jw-msft, line 123: why adding back tick back? revert this file to the commit before

nihemstr: I have no idea why this was changed. Checking files again to see whats going on

Nox-MSFT (approved):

jw-msft (approved):

PR #361, opened 2023-01-24
Changed updateType telemetry parsing to match tracing

nihemstr:
- Added clarification for the swupdate trace as per bug: https://microsoft.visualstudio.com/OS/_queries/edit/43011595/?triage=true

### DIFF
--- a/src/extensions/step_handlers/swupdate_handler/src/swupdate_handler.cpp
+++ b/src/extensions/step_handlers/swupdate_handler/src/swupdate_handler.cpp
@@ -143,7 +143,8 @@ ADUC_Result SWUpdateHandlerImpl::Download(const tagADUC_WorkflowData* workflowDa
     updateTypeOk = ADUC_ParseUpdateType(updateType, &updateName, &updateTypeVersion);
     if (!updateTypeOk)
     {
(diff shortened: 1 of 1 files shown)


src/extensions/step_handlers/swupdate_handler/src/swupdate_handler.cpp
```diff
@@ -143,7 +143,8 @@ ADUC_Result SWUpdateHandlerImpl::Download(const tagADUC_WorkflowData* workflowDa
     updateTypeOk = ADUC_ParseUpdateType(updateType, &updateName, &updateTypeVersion);
     if (!updateTypeOk)
     {
-        Log_Error("SWUpdate packages download failed. Unknown Handler Version (UpdateDateType:%s)", updateType);
+        Log_Error(
```
JeffMill, line 146: why not output updateType in trace? seems like that's relevant?

JeffMill (approved):
approved with comment.

src/extensions/step_handlers/swupdate_handler/src/swupdate_handler.cpp
nihemstr: There's only one way the updateType can't be parsed - when it's not following the delineated path. Plus there's the option that updateType is not properly allocated/handled. Hence why I removed the printout.

PR #353, opened 2023-01-19
No alloc FileEntity from workflow_get_update_file

jw-msft:
workflow_get_update_file and workflow_get_update_file_by_name unnecessarily allocate the FileEntity for the out param. Fixed to just pass in address of stack variable instead to avoid fragmenting the heap.

PR #362, opened 2023-01-24
No alloc FileEntity from workflow_get_update_file (#353)

JeffMill:
* no alloc of FileEntity from workflow_get_update_file

* remove workflow_free_file_entity and nonalloc out for workflow_get_step_detached_manifest_file


src/extensions/update_manifest_handlers/steps_handler/src/steps_handler.cpp
```diff
@@ -127,6 +122,9 @@ ADUC_Result PrepareStepsWorkflowDataObject(ADUC_WorkflowHandle handle)
             }
             else
             {
+                ADUC_FileEntity entity;
```
jw-msft, line 125: would it better to have at outermost scope so that ADUC_FileEntity_Uninit could be called at done: label like in develop branch?

PR #364, opened 2023-01-25
SHARED -> MODULE to avoid linking to extensions

jw-msft:
as per [CMake add_library doc](https://cmake.org/cmake/help/latest/command/add_library.html#command:add_library):
`SHARED libraries are linked dynamically and loaded at runtime. MODULE libraries are plugins that are not linked into other targets but may be loaded dynamically at runtime using dlopen-like functionality`

The intention of extensions is to be plugins, not to allow other targets to directly link with them.  This only affects non-windows platforms.


JeffMill (approved):
i already fixed this in wiot, but ill CP this for consistency

PR #371, opened 2023-02-01
Remove sleep; Clean up d2c_messaging_ut nanosleep calls.

JeffMill:


src/utils/d2c_messaging/tests/d2c_messaging_ut.cpp
```diff
@@ -186,35 +192,15 @@ void* mock_msg_process_thread_routine(void* context)
     // Wait before response.
     if (g_cloudBehavior[g_cloudBehaviorIndex].delayBeforeResponseMS > 999)
     {
-        ADUCPAL_sleep(
-            (static_cast<unsigned int>(g_cloudBehavior[g_cloudBehaviorIndex].delayBeforeResponseMS + 500) / 1000));
+        timespec t;
```
Nox-MSFT, line 195: For linux os, we can't call nanosleep here if value (ms) > 999.

Per man-page: 

       The value of the nanoseconds field must be in the range 0 to
       999999999.

Nox-MSFT (changes requested):

yuvalg-MSFT (approved):

src/utils/d2c_messaging/tests/d2c_messaging_ut.cpp
Nox-MSFT: (When not define ADUCPAL_USE_PAL, ADUCPAL_nanosleep => nanosleep)

JeffMill: good catch.  darn it.

JeffMill: Weird. I didn't see that here: https://linux.die.net/man/3/nanosleep

JeffMill: this works, however:

        ts.tv_sec = 3;
        ts.tv_nsec = 0;
        nanosleep(&ts, NULL);

I'll fix my set_timespec_ms() helper.

Nox-MSFT (approved):

PR #372, opened 2023-02-02
remove mkstemp, replace with system utils method. remove related aducpal apis.

JeffMill:


src/utils/system_utils/src/system_utils.c
```diff
@@ -67,6 +68,52 @@ const char* ADUC_SystemUtils_GetTemporaryPathName()
...
+    {
+        tmpl[0] = '\0';
+        errno = EINVAL;
+        return tmpl;
+    }
+
+    // Start of XXXXXX
+    char* Xs = tmpl + (len - 6);
+
+    // Get a randomish number from real time clock.
```
jw-msft, line 96: why not use srand()/rand()?  Don't see a compelling reason to roll one's own PRNG.  If for production use, it could use /dev/random or /dev/urandom on *nix and {Rtl,Crypt}GenRandom on win32, or implement in c++ using <random>

src/utils/system_utils/src/system_utils.c
```diff
@@ -67,6 +68,52 @@ const char* ADUC_SystemUtils_GetTemporaryPathName()
...
+ *
+ * However, for unit tests, it can be useful, so it's highly recommended that this method is only
+ * used for unit tests.
+ *
+ * @param tmpl The last six characters XXXXXX are replaced with a string that makes the filename unique.
+ * @return Returns @param tmpl.
+ */
+char* ADUC_SystemUtils_MkTemp(char* tmpl)
+{
+    const size_t len = strlen(tmpl);
```
jw-msft, line 85: it'd be safer if char len of the template string were passed in.

src/utils/system_utils/src/system_utils.c
```diff
@@ -67,6 +68,52 @@ const char* ADUC_SystemUtils_GetTemporaryPathName()
     return env;
 }
 
+/**
+ * @brief mktemp() implementation
+ *
+ * mktemp() is generally considered unsafe as the temporary names can be easy to guess, and there are
+ * a limited number of combinations.
+ *
+ * However, for unit tests, it can be useful, so it's highly recommended that this method is only
```
jw-msft, line 77: If putting it in system_util it would eventually get used despite the comments.  Please move this into test_utils instead if only meant for test scenarios.

jw-msft (commented):
some comments

src/utils/system_utils/src/system_utils.c
JeffMill: that's not how mktemp works, so I didn't want to deviate, but I agree in principle.

src/utils/system_utils/src/system_utils.c
JeffMill: i started with that and but rand likes to return values like 1000 and 23, which isn't enough digits for my needs. Maybe I was doing something wrong?

PR #373, opened 2023-02-03
Fix SWUpdate handler V2 device reboot and agent restart issue.

Nox-MSFT:


src/extensions/step_handlers/swupdate_handler_v2/src/swupdate_handler_v2.cpp
```diff
@@ -262,12 +262,36 @@ ADUC_Result SWUpdateHandler_PerformAction(
...
 
 done:
-    if (IsAducResultCodeFailure(result.ResultCode) && workflowData->WorkflowHandle != nullptr)
+    if (IsAducResultCodeFailure(result.ResultCode))
     {
         workflow_set_result(workflowData->WorkflowHandle, result);
         workflow_set_state(workflowData->WorkflowHandle, ADUCITF_State_Failed);
     }
 
+    // Note: the handler must request a system reboot or agent restart if required.
```
JeffMill, line 271: this method is already  pretty big, and these results only apply to install and result.  should n't these be checked in the install and apply handlers specifically?

also, when did we add reboot after install?  I believe it was originally only for apply which makes more sense to me.  even if a handler doesn't have work to do in apply , it can still return reboot from that,

JeffMill: interestingly, it looks like this used to be done in apply like i suggest, but you've moved it here?

JeffMill (changes requested):

src/extensions/step_handlers/swupdate_handler_v2/src/swupdate_handler_v2.cpp
Nox-MSFT: Customer can choose to request the agent to reboot or restart even during install phase. IMO, in the future, we need to decouple any workflow-related request from the resultCode, and provide the APIs for the handlers, or backing scripts to make a request to the agent core directly.

As for this change, I'm making this behavior consistent with the Script Handler.

Nox-MSFT: (FWIW, we can even completely deprecate SWUpdate Handler and SWUpdate handler V2, since the same update scenario can be accomplished using Script Handler)

JeffMill: your replies are related to the second half of my comments.  My first comment still applies however:

*this method is already pretty big, and these results only apply to install and result. shouldn't these be checked in the install and apply handlers specifically?*

JeffMill (changes requested):

JeffMill (dismissed):

src/extensions/step_handlers/swupdate_handler_v2/src/swupdate_handler_v2.cpp
Nox-MSFT: Fixed

src/extensions/step_handlers/swupdate_handler_v2/tests/testdata/adu-yocto-ab-rootfs-update/payloads/example-a-b-update.sh
```diff
@@ -657,14 +659,18 @@ ApplyUpdate() {
...
+        log_info "restart_to_apply=$restart_to_apply"
+        log_info "restart_agent_to_apply=$restart_agent_to_apply"
+
+        if [[ ${restart_to_apply} == "yes" ]]; then
+            log_info "Returning ADUC_Result_Apply_RequiredImmediateReboot(705)"
             resultCode=705
-        elif [ "$restart_agent_to_apply" == "yes" ]; then
-            # ADUC_Result_Apply_RequiredImmediateAgentRestart = 707
+        elif [[ ${restart_agent_to_apply} == "yes" ]]; then
+            log_info "Retrurning ADUC_Result_Apply_RequiredImmediateAgentRestart(707)"
```
jw-msft, line 670: typo: Retrurning

jw-msft (approved):

PR #374, opened 2023-02-04
Set default DO ref to main

shiyi-peng:
To better align DU and DO development, and to resolve 
CMake Error in sdk-cpp/CMakeLists.txt:
  Imported target "Boost::filesystem" includes non-existent path

    "/include"

  in its INTERFACE_INCLUDE_DIRECTORIES.

### DIFF
--- a/scripts/install-deps.sh
+++ b/scripts/install-deps.sh
@@ -72,7 +72,7 @@ install_githooks=false
 
 du_test_data_dir_path="/tmp/adu/"
 # DO Deps
-default_do_ref=develop
+default_do_ref=v1.0.1
 install_do=false
 do_ref=$default_do_ref
 
--- a/scripts/install-deps.sh
+++ b/scripts/install-deps.sh
@@ -72,7 +72,7 @@ install_githooks=false
 
 du_test_data_dir_path="/tmp/adu/"
 # DO Deps
-default_do_ref=main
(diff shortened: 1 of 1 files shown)


scripts/install-deps.sh
```diff
@@ -72,7 +72,7 @@ install_githooks=false
 
 du_test_data_dir_path="/tmp/adu/"
 # DO Deps
-default_do_ref=v1.0.0
```
JeffMill, line 75: for develop dont we want to lock to do's develop branch?

also, this doesn't fix our release which is still broken for ubuntu 20 by default.

JeffMill (changes requested):

scripts/install-deps.sh
```diff
@@ -72,7 +72,7 @@ install_githooks=false
...
+default_do_ref=develop
```
jw-msft, line 75: for develop, it can be develop.  If destined for main, it should be either a commit hash or tag so `<commit hash>` is better than `main`, which is a potential supply chain issue since it would always be pulling down latest main, which always has potential to break things (or worse).

jw-msft (approved):

yuvalg-MSFT (approved):

yuvalg-MSFT (approved):

PR #381, opened 2023-02-11
Install gcc-10 for Debian 11 Bullseye for RPi4

jw-msft:


scripts/install-deps.sh
```diff
@@ -166,6 +166,8 @@ do_install_aduc_packages() {
     OS=$(lsb_release --short --id)
     if [[ $OS == "debian" && $VER == "9" ]]; then
         $SUDO apt-get install --yes gcc-6 g++-6 || return
+    elif [[ $OS == "Debian" && $VER == "11" ]]; then
```
JeffMill, line 169: nit: combine 167 and 169

```
if OS = debian
  if ver = 9
  else ver = 11
  else  fail?
else
```

JeffMill (approved):
approved with suggestion.

scripts/install-deps.sh
jw-msft: The current else is serving as Debian 10 and ubuntu catch all atm. Will refactor to be more explicit across all supported ubuntu and debian later.

PR #384, opened 2023-02-15
Merge branch 'fix-independent-code' of github.com:neonsoftware/iot-hub-device-update

jw-msft:
Merge and Conflict Resolution of following PR into develop: https://github.com/Azure/iot-hub-device-update/pull/181

Co-authored-by: neonsw <neonsw@neonsw.com>

PR #385, opened 2023-02-16
fix: cmake misses some POSITION_INDEPENDENT_CODE settings

neonsoftware:
Other libraries in this project set POSITION_INDEPENDENT_CODE, ex. workflow_utils, and *depend* on these libraries.

Therefore, they need to be position indepent as well.

Fixes compilation issues seen first in OpenWRT for device update v0.8.0.


neonsoftware:
An expanded version of #384

jw-msft (approved):

PR #386, opened 2023-02-16
fix: add missing explicit link to atomic library

neonsoftware:
With some compiler/libc scenarios, the _link to atomic is implicit_.
With others, instead, such link is required to be _explicit_.

One example is OpenWRT 21 + gcc + musl.

This change should, normally, not affect scenarios where it is implicit.


src/agent/CMakeLists.txt
```diff
@@ -95,6 +95,7 @@ target_link_libraries (
             aduc::permission_utils
             aduc::pnp_helper
             aduc::system_utils
+            atomic
```
jw-msft, line 98: Based on instances of #include <atomic>, I would expect the modeling of direct dependencies in CMake to be the following instead of as private target link library in agent.
(where PRIVATE is for impl and PUBLIC is for header+impl (and INTERFACE for header-only))

target_link_libraries

- Add to PRIVATE in src/extensions/content_downloaders/deliveryoptimization_downloader/CMakeLists.txt
- Add to PUBLIC in src/platform_layers/linux_platform_layer/CMakeLists.txt
- Add to PRIVATE in src/platform_layers/linux_platform_layer/tests/CMakeLists.txt

```sh
$ rg '<atomic>'
src/extensions/content_downloaders/deliveryoptimization_downloader/deliveryoptimization_content_downloader.cpp
15:#include <atomic>

src/platform_layers/linux_platform_layer/src/linux_adu_core_impl.hpp
11:#include <atomic>

src/platform_layers/linux_platform_layer/tests/mock_do_download.hpp
12:#include <atomic>
```

jw-msft (changes requested):

jw-msft (approved):
Thanks for fixing this.  Approved.

PR #390, opened 2023-02-16
Remove remaining usages of _Bool from generated error code.

neonsoftware:
All other previous usages of `_Bool` were already removed in PR #268.

These seem to be the last two left.

This patch is actually necessary in some compiler scenarios using Musl.

### DIFF
--- a/scripts/error_code_generator_defs/error_code_defs_generator.py
+++ b/scripts/error_code_generator_defs/error_code_defs_generator.py
@@ -46,7 +46,7 @@ typedef enum tagADUC_GeneralResult
 /**
 * @brief Determines if a result code is succeeded.
 */
-static inline _Bool IsAducResultCodeSuccess(const ADUC_Result_t resultCode)
+static inline bool IsAducResultCodeSuccess(const ADUC_Result_t resultCode)
 {{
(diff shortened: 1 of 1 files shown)


jw-msft (approved):

jw-msft:
PRs from forks currently fail in pr pipeline that will be addressed later, so I cherry-picked into the following branch so we can run the pr pipeline on it:
https://github.com/Azure/iot-hub-device-update/tree/user/jw-msft/neongenbool

The results are running here (unfortunately failure details would not be accessible externally but I can relay any, although I don't expect it for these):
https://dev.azure.com/azure-device-update/adu-linux-client/_build/results?buildId=11921&view=results

jw-msft:
> PRs from forks currently fail in pr pipeline that will be addressed later, so I cherry-picked into the following branch so we can run the pr pipeline on it: https://github.com/Azure/iot-hub-device-update/tree/user/jw-msft/neongenbool
> 
> The results are running here (unfortunately failure details would not be accessible externally but I can relay any, although I don't expect it for these): https://dev.azure.com/azure-device-update/adu-linux-client/_build/results?buildId=11921&view=results

and it succeeded as expected.

PR #109, opened 2021-12-10
Split adushell consts to util library

mdavis777:
This will allow for multiple handlers to be included at the same time

PR #403, opened 2023-02-22
Split adushell consts to util library

jw-msft:
Ported from PR #109

Co-authored-by: Mike Davis <mdavis@witekio.com>


JeffMill:
Original PR says:

`This will allow for multiple handlers to be included at the same time`

Do you understand what the issue is here?  Is he not able to build with two handlers?

PR #412, opened 2023-03-01
Add decrypted file handling

jw-msft:
- Add decrypted file handling download logic 
- add download_utils to build content downloader fullFilePath


jw-msft:
shelving this for now

PR #416, opened 2023-03-03
Change retryTimeout to maxTimeoutInSeconds

jw-msft:
Renames in content downloader download to be less confusing and match the reality of the content downloader contract.
- Rename DO_RETRY_TIMEOUT_DEFAULT to CONTENT_DOWNLOADER_MAX_TIMEOUT_IN_SECONDS_DEFAULT.
- Rename retryTimeout to maxTimeoutInSeconds

https://github.com/microsoft/do-client/blob/8df51fc107568b6c3cc2ea205676efe726ea7ddb/sdk-cpp/include/do_download.h#L49
Argument in DO API is timeoutSecs:
static std::error_code download_url_to_path(const std::string& uri, const std::string& downloadFilePath, const std::atomic_bool& isCancelled, std::chrono::seconds **timeoutSecs** = std::chrono::hours(24)) noexcept;

Tracing that code down, it ends up using it to calculate and `endTime` for the polling loop to exit and give up on the download here:
https://github.com/microsoft/do-client/blob/develop/sdk-cpp/src/do_download.cpp#L103


src/extensions/extension_manager/inc/aduc/extension_manager.hpp
```diff
@@ -21,8 +21,8 @@
 #include <string>
 #include <unordered_map>
 
-// Default DO retry timeout is 24 hours.
-#define DO_RETRY_TIMEOUT_DEFAULT (60 * 60 * 24)
+// Default Content Downloader Maximum Timeout in seconds before the downloader gives up trying to download.
```
JeffMill, line 24: do we really want to make the default timeout 24 hours?  that's an incredibly long time unless someone is over a satellite connection or something.

src/extensions/extension_manager/inc/aduc/extension_manager_download_options.h
```diff
@@ -3,7 +3,8 @@
 
 typedef struct tagADUC_ExtensionManager_Download_Options
 {
-    unsigned int retryTimeout;
+    unsigned int
+        maxTimeoutInSeconds; /**< The maximum number of seconds the content downloader should try to keep downloading. */
```
JeffMill, line 7: is seconds the best unit to use here?  do we feel a customer would decide to use 63 seconds vs 127 seconds?

Seems like minutes is more logical.

src/extensions/extension_manager/src/extension_manager.cpp
```diff
@@ -46,7 +46,7 @@ using ADUC::StringUtils::cstr_wrapper;
 
 EXTERN_C_BEGIN
 ExtensionManager_Download_Options Default_ExtensionManager_Download_Options = {
-    .retryTimeout = 60 * 60 * 24 /* default : 24 hour */,
+    .maxTimeoutInSeconds = 60 * 60 * 24 /* default : 24 hour */,
 };
```
JeffMill, line 50: as above.  24 hours?

JeffMill (changes requested):

src/extensions/extension_manager/inc/aduc/extension_manager.hpp
jw-msft: Made it 8 hours and in minutes (within extension manager domain).  It still has to stay as seconds units b/c that's the content download interface contract, so I convert it.

src/extensions/extension_manager/inc/aduc/extension_manager_download_options.h
```diff
@@ -1,9 +1,13 @@
 #ifndef ADUC_EXTENSION_MANAGER_DOWNLOAD_OPTIONS_H
 #define ADUC_EXTENSION_MANAGER_DOWNLOAD_OPTIONS_H
 
+// Default Content Downloader Download Timeout of 8 hours.
+#define CONTENT_DOWNLOADER_MAX_TIMEOUT_IN_MINUTES_DEFAULT 8 * 60
```
JeffMill, line 5: nit? `(8 * 60)`

JeffMill (approved):

JeffMill (approved):

PR #420, opened 2023-03-08
Ubuntu Core Snap - AzIoTIdentity integration

shiyi-peng:
E2E test with AIS Snap works, pending some modifications on Unit Test & merge with the changes on DO snap changes


src/utils/eis_utils/src/eis_coms.c
```diff
@@ -32,17 +32,17 @@
 /**
  * @brief Unix Domain Socket (UDS) for the Identity Service API
  */
-#define EIS_UDS_IDENTITY_SOCKET_PATH "/run/aziot/identityd.sock"
+#define EIS_UDS_IDENTITY_SOCKET_PATH "/var/sockets/aziot/identityd.sock"
```
Nox-MSFT, line 35: Make this a compile time value. We need to continue to support non-snap version.

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -418,6 +424,82 @@ EISErr BuildSasCertConnectionString(
...
+}
+
+/**
+ * @brief Reads EIS Identities information from a EIS data file
+ * @details Reads EIS Identity from a file, first line being the identity response, 
+ * and the second line being the EIS Error Code.
+ * @param[in,out] identityResponseBuffer the pointer to Identity response string
+ * @returns a value of EISErr
+ */
+EISErr EISIdentitiesFileReader(char** identityResponseBuffer)
```
Nox-MSFT, line 467: NULL check the identityResponseBuffer before referencing it.

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -468,208 +550,241 @@ EISUtilityResult RequestConnectionStringFromEISWithExpiry(
     JSON_Value* certResponseJson = NULL;
     char* certString = NULL;
 
-    EISErr identityResult = RequestIdentitiesFromEIS(timeoutMS, &identityResponseStr);
+    printf("Parent process (PID %d) is about to fork a child process...\n", getpid());
```
Nox-MSFT, line 553: In this function, consider using named-pipe for IPC between parent process and child process. I think the codes might be simpler.

Nox-MSFT (changes requested):

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -468,208 +550,241 @@ EISUtilityResult RequestConnectionStringFromEISWithExpiry(
...
+            result = BuildSharedAccessSignature(
+                resourceUri, keyHandle, expirySecsSinceEpoch, timeoutMS, &sharedSignatureStr);
+
+            if (result.err != EISErr_Ok)
+            {
+                goto done;
+            }
+
+            result.err = BuildSasTokenConnectionString(
+                hubName, deviceId, moduleId, connType, sharedSignatureStr, gatewayHostName, &connectionStr);
```
nihemstr, line 708: Where are you requesting the construction of the sharedSignature? I don't see a call anywhere here to the key service?

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -468,208 +550,241 @@ EISUtilityResult RequestConnectionStringFromEISWithExpiry(
...
+            const char* certId = json_object_get_string(authJson, EIS_IDENTITY_RESP_AUTH_CERTID_FIELD);
+
+            if (certId == NULL)
+            {
+                result.err = EISErr_InvalidJsonRespErr;
+                result.service = EISService_IdentityService;
+                goto done;
+            }
+
+            EISErr certResult = RequestCertificateFromEIS(certId, timeoutMS, &certResponseStr);
```
nihemstr, line 734: The UID of this process needs to be the same as the one in the principle (ais_user I believe) in order to make this call. 

It looks like the file reader may be a separate user?

packages/debian/postinst
```diff
@@ -239,22 +242,22 @@ register_reference_extensions() {
 
 register_adu_user_and_process_with_eis() {
         if getent group "aziotid" > /dev/null; then
-            usermod -aG aziotid $adu_user
+            usermod -aG aziotid $ais_user
```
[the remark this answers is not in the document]
Nox-MSFT: @nihemstr, On Ubuntu Core, we can't create arbitrary user and group. The special user and group (snap_aziot_du) has been white-listed and created specifically for DU Agent snap. I discussed this off-line with @shiyi-peng and we will revert this change since postinst script is intended to be used during Dabien package installation only.

For DU Snap, the principal file will be created during Snap installation using 'configure  hook'

src/utils/eis_utils/src/eis_utils.c
shiyi-peng: not changing anything on this part of the function. I think git just messed everything up! :(

src/utils/eis_utils/src/eis_utils.c
shiyi-peng: yeah it is a separate user... oh no

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -418,36 +425,109 @@ EISErr BuildSasCertConnectionString(
...
+        write(pipeFd, identityResponseStr, strlen(identityResponseStr));
+        write(pipeFd, "\n", 1);
+
+        char identityResultStr[32];
+        snprintf(identityResultStr, sizeof(identityResultStr), "%d\n", identityResult);
+        write(pipeFd, identityResultStr, strlen(identityResultStr));
+
+        close(pipeFd);
+
+        printf("Connection data written to named pipe %s.\n", pipePath);
```
Nox-MSFT, line 461: Let's not print the pipePath here. Should keep the path internal to our app.

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -418,36 +425,109 @@ EISErr BuildSasCertConnectionString(
...
-// External Functions
-//
-
 /**
- * @brief Creates a connection string using the provisioned data within EIS
- * @details Calls into the EIS Identity and Keyservice to create a SharedAccessSignature which is then used
- * to create the connection string, Caller is required to call free() to deallocate the connection string
- * @param[in] expirySecsSinceEpoch the expiration time in seconds since the epoch for the token in the connection string
+ * @brief Writing EIS Identities information to a EIS data file
+ * @details Calls into the EIS Identity and get the identities response, saves to a file
```
Nox-MSFT, line 430: Please update the comment to match the new design.

CMakeLists.txt
```diff
@@ -328,6 +328,11 @@ set (
     "adu"
     CACHE STRING "The effective group name for adu-shell process.")
 
+set (
+    EIS_PIPE_PATH
+    "/tmp/eis_pipe"
```
Nox-MSFT, line 333: Since this will contains device connection info, let's use the path that accessible by snap_aziot_du group.
Perhaps /var/lib/adu/<eis_response_file_name> ?

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -701,3 +770,99 @@ EISUtilityResult RequestConnectionStringFromEISWithExpiry(
...
+
+
+    printf("Parent process (PID %d) is about to fork a child process...\n", getpid());
+
+    // Create a named pipe (FIFO)
+    if (mkfifo(EIS_PIPE_PATH, 0666) == -1) {
+        result.err = EISErr_NamedPipeFailure;
+        goto done;
+    }
+
```
Nox-MSFT, line 814: Consider setting the file owner/group to snap_aziot_du or adu.

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -418,36 +425,109 @@ EISErr BuildSasCertConnectionString(
...
     {
-        result.err = EISErr_InvalidArg;
-        return result;
+        int pipeFd = open(pipePath, O_WRONLY);
+        if (pipeFd < 0)
+        {
+            retCode = EISErr_NamedPipeFailure;
+            goto done;
+        }
+        // dprintf(pipeFd, "%s\n", identityResponseStr);
```
Nox-MSFT, line 450: remove commented lines.

src/utils/eis_utils/src/eis_err.c
```diff
@@ -58,6 +58,8 @@ const char* EISErr_ErrToString(EISErr eisErr)
         return "EISErr_ContentAllocErr";
     case EISErr_InvalidJsonRespErr:
         return "EISErr_InvalidJsonRespErr";
+    case EISErr_NamedPipeFailure:
```
Nox-MSFT, line 61: Consider more descriptive name here. e.g., EISErr_RespNamedPipeInitErr, or something along this line. In case we want to add read or write error codes in the future.

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -418,36 +425,109 @@ EISErr BuildSasCertConnectionString(
...
+    {
+        retCode = EISErr_ConnErr;
+    }
+done:
+    return retCode;
+}
 
-    if (expirySecsSinceEpoch <= time(NULL) || timeoutMS == 0)
+/**
+ * @brief Reads EIS Identities information from a EIS data file
```
Nox-MSFT, line 472: Missing comment for pipePath.

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -418,36 +425,109 @@ EISErr BuildSasCertConnectionString(
...
+    EISErr result=0;
+    int pipeFd = open(pipePath, O_RDONLY);
+    if (pipeFd < 0)
     {
-        result.err = EISErr_InvalidArg;
+        *childErrno = errno;
         return result;
     }
 
+    *identityResponseBuffer = malloc(1024 * sizeof(char));
```
Nox-MSFT, line 489: Instead of hard-coding 1024, consider #define or declare constant var at the top of the function and use it throughout the function.

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -418,36 +425,109 @@ EISErr BuildSasCertConnectionString(
...
+ */
+EISErr EISIdentitiesPipeReader(const char* pipePath, char** identityResponseBuffer, int* childErrno)
+{
+    EISErr result=0;
+    int pipeFd = open(pipePath, O_RDONLY);
+    if (pipeFd < 0)
     {
-        result.err = EISErr_InvalidArg;
+        *childErrno = errno;
         return result;
```
Nox-MSFT, line 486: Should we return EISErr_***pipeFailure here?

Also, throughout this function, would be great if we (from log) can diagnose each failure conditions based on combination of childErrno + returned EISErr. Try to return the most descriptive EISErr + errno. Consider add new EISErr_ if needed.

Nox-MSFT (changes requested):

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -418,36 +425,109 @@ EISErr BuildSasCertConnectionString(
...
-        return result;
+        int pipeFd = open(pipePath, O_WRONLY);
+        if (pipeFd < 0)
+        {
+            retCode = EISErr_NamedPipeFailure;
+            goto done;
+        }
+        // dprintf(pipeFd, "%s\n", identityResponseStr);
+        // dprintf(pipeFd, "%d\n", identityResult);
+        write(pipeFd, identityResponseStr, strlen(identityResponseStr));
```
jw-msft, line 452: why not use the higher-level fprintf followed by a fflush so you can use the string format?

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -418,36 +425,109 @@ EISErr BuildSasCertConnectionString(
...
+        write(pipeFd, "\n", 1);
```
jw-msft, line 453: would need to check return value for -1 and then check errno foreach write call

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -418,36 +425,109 @@ EISErr BuildSasCertConnectionString(
...
+
+        char identityResultStr[32];
+        snprintf(identityResultStr, sizeof(identityResultStr), "%d\n", identityResult);
```
jw-msft, line 456: this would all be simpler with just a single fprintf call

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -418,36 +425,109 @@ EISErr BuildSasCertConnectionString(
...
         return result;
     }
 
+    *identityResponseBuffer = malloc(1024 * sizeof(char));
+    if (*identityResponseBuffer == NULL) {
+        *childErrno = errno;
+        goto done;
+    }
+    
+    if (read(pipeFd, *identityResponseBuffer, 1024) < 0) {
```
jw-msft, line 495: This can all be simplified with fopen + fscanf as you won't need to manually do string -> int conversion.

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -701,3 +770,99 @@ EISUtilityResult RequestConnectionStringFromEISWithExpiry(
...
+    if (expirySecsSinceEpoch <= time(NULL) || timeoutMS == 0)
+    {
+        result.err = EISErr_InvalidArg;
+        return result;
+    }
+
+    char* identityResponseStr = NULL;
+
+
+    printf("Parent process (PID %d) is about to fork a child process...\n", getpid());
```
jw-msft, line 807: Should this be a debug logging trace instead?

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -701,3 +770,99 @@ EISUtilityResult RequestConnectionStringFromEISWithExpiry(
...
+            result.service = EISService_IdentityService;
+            result.err = EISResult;
+            errno = childErrno; // Set the errno to the value received from the child process
+            goto done;
+        }
+
+        result = ProcessIdentityResponse(provisioningInfo, identityResponseStr, expirySecsSinceEpoch, timeoutMS);
+
+    }
+done:
```
jw-msft, line 865: should set ResultCode to success right before done label

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -701,3 +770,99 @@ EISUtilityResult RequestConnectionStringFromEISWithExpiry(
...
+
+        if (EISResult != EISErr_Ok)
+        {
+            result.service = EISService_IdentityService;
+            result.err = EISResult;
+            errno = childErrno; // Set the errno to the value received from the child process
+            goto done;
+        }
+
+        result = ProcessIdentityResponse(provisioningInfo, identityResponseStr, expirySecsSinceEpoch, timeoutMS);
```
jw-msft, line 862: need to check result and goto done if failed

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -418,36 +425,109 @@ EISErr BuildSasCertConnectionString(
...
+        goto done;
+    }
+    
+    if (read(pipeFd, *identityResponseBuffer, 1024) < 0) {
+        *childErrno = errno;
+        free(*identityResponseBuffer);
+        goto done;
+    }
+
+    char readidentityResult[1024];
```
jw-msft, line 501: please declare variables at beginning of the block, also for line 508. It's compiling because of gnu compiler extensions, which could cause compat issues with other compilers in the future (msvc, clang, etc).  We can fix it overall by adding -pedantic cflag.

src/utils/eis_utils/src/eis_utils.c
```diff
@@ -418,36 +425,109 @@ EISErr BuildSasCertConnectionString(
...
 
+    *identityResponseBuffer = malloc(1024 * sizeof(char));
+    if (*identityResponseBuffer == NULL) {
+        *childErrno = errno;
+        goto done;
+    }
+    
+    if (read(pipeFd, *identityResponseBuffer, 1024) < 0) {
+        *childErrno = errno;
+        free(*identityResponseBuffer);
```
jw-msft, line 497: put the free after done label if err. I'd suggest only setting out params right before or right after done label.

src/utils/eis_utils/src/eis_utils.c
jw-msft, line 489: Should use stack char[] and only malloc and create output buffer when committing upon full success; otherwise, you have to malloc+free unnecessarily when there's an error.

jw-msft (changes requested):

jw-msft:
Change PR and commit title to not include internal work item task numbers as they are not useful to the commit log and external contributors. PR + commit links could be added to internal work items, but not the other way around.

CMakeLists.txt
shiyi-peng: "/var/lib/adu/eis_response_pipe" works?

src/utils/eis_utils/src/eis_utils.c
shiyi-peng: But I think it automatically goes to done because it is end of function?

src/utils/eis_utils/src/eis_utils.c
shiyi-peng: shouldn't it return whatever result ProcessIdentityResponse returns?

Issue #367, opened 2023-01-30
DU Agent does not update Twin when credentials are refreshed


agreentree-dmc:
## Expected Behavior
Using x509 certificates as Device Authentication Type. When certificates are periodically rotated by Azure Identity Service, the DU Agent re-establishes it's connection with the IoT Hub.  The DU Agent **_should_** refresh Module Identity Twin when connection is re-established and remain in healthy state without needing to restart the service.

## Current Behavior
Using x509 certificates as Device Authentication Type. When certificates are periodically rotated by Azure Identity Service, the DU Agent re-establishes it's connection with the IoT Hub.  The DU Agent **_does not_** update the Module Identity Twin which leaves the device in an unhealthy, not updatable state. Service must be restarted to get DU Agent back to healthy state.

## Steps to Reproduce
1. Install Azure Identity Service (AIS) 1.4.0.
2. Configure AIS to use self-signed certificates issued from an EST Server (as described in this tutorial: https://learn.microsoft.com/en-us/azure/iot-edge/tutorial-configure-est-server?view=iotedge-1.4). Set the `threshold` property under `[provisioning.attestation.identity_cert.auto_renew]` in the aziot config.toml file to be 10% to observe frequent certificate rotations (as described in suggestion 2 at the end of the 'Test Certificate Renewal' section of the EST Server Tutorial). 
3. Configure AIS to use 'DPS provisioning with X.509 certificate' method.
4. Install ADU 1.0.1 and configure it to obtain credentials from AIS. 
5. Observer the Module Identity Twin in portal just after the DU agent successfully starts for the first time.
6. Wait for certificate renewal. 
7. Observe the Module Identity Twin in portal after the certificate has rotated.

### Device Information
* Host OS: NI Linux Real-Time 9.1 (hardknott)
* Architecture: amd64
* Provisioning Method: DPS Provisioning with X.509 certificate
* DU Agent Version: 1.0.1

## Logs
Main log that I think shows the behavior when the device certificate is refreshed and the connection is re-established (towards the bottom): 
[du-agent.20230127-162704.log](https://github.com/Azure/iot-hub-device-update/files/10539669/du-agent.20230127-162704.log)

Zip of all logs in the /var/log/adu folder for reference: 
[du_logs.zip](https://github.com/Azure/iot-hub-device-update/files/10539677/du_logs.zip)

## Additional Information
_Screenshots after certificate rotation_
DU Agent connection status:  
[...]

PR #422, opened 2023-03-09
added handling to restart adu in case of "unauthenticated:no network"

JoergZeidler:
Hi iot-hub-device-update Team,

Regarding issue 
https://github.com/Azure/iot-hub-device-update/issues/367 
and 
https://github.com/Azure/iot-hub-device-update/issues/212

i added a new handling for the communication handler in case of "unauthenticated " state with reason "no network".
I saw on multiple logs on my side that in case "no network" the module twin is empty for the reported properties items.
That could happen if the identity service re-provision the device. This PR should fix that behaviour.

new handling:
1) try to restart agent to prevent empty reported properties in module twin  
2) if restart agent could not be performed, wait for at least 5 minutes to retry.

### DIFF
--- a/src/communication_managers/iothub_communication_manager/src/iothub_communication_manager.c
+++ b/src/communication_managers/iothub_communication_manager/src/iothub_communication_manager.c
@@ -5,6 +5,7 @@
  * @copyright Copyright (c) Microsoft Corporation.
  * Licensed under the MIT License.
  */
+#include "aduc/adu_core_export_helpers.h" // ADUC_MethodCall_RestartAgent
 #include "aduc/adu_types.h"
 #include "aduc/client_handle_helper.h"
 #include "aduc/config_utils.h"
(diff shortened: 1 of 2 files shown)


src/communication_managers/iothub_communication_manager/src/iothub_communication_manager.c
```diff
@@ -725,9 +726,14 @@ static void Connection_Maintenance()
             additionalDelayInSeconds = TIME_SPAN_FIVE_MINUTES_IN_SECONDS;
             break;
         case IOTHUB_CLIENT_CONNECTION_NO_NETWORK:
-            // Could be transient error, wait for at least 5 minutes to retry.
+            // 1) try to restart agent to prevent empty reported properties in module twin 
+            // 2) if restart agent could not be performed, wait for at least 5 minutes to retry.
             Log_Error("No network.");
-            additionalDelayInSeconds = TIME_SPAN_FIVE_MINUTES_IN_SECONDS;
+            if (ADUC_MethodCall_RestartAgent() != 0)
```
nihemstr, line 732: NoNetwork has a lot of meanings in the iot-hub-c-sdk. I'm ok with a hard restart as an option, but do we have a max retries for restarting the daemon after a failure? I know we can rely on the exponential back off of the daemon for the retry stuff but what happens if we hit our (5 I believe) max retries? Do we have a restart story?

nihemstr (approved):
Approved with comments

src/communication_managers/iothub_communication_manager/src/iothub_communication_manager.c
JoergZeidler: the default value of the deviceupdate-agent.service is to restart always after a timeout of 5s. https://github.com/JoergZeidler/iot-hub-device-update-azure/blob/e0c7f3b83292479c0f4666fbaf3e53d8303c7f66/daemon/deviceupdate-agent.service#L8

PR #425, opened 2023-03-13
[Test Scenario] Add Agent Startup and Basic Function Verification Test

shiyi-peng:


yuvalg-MSFT (approved):
need to make sure everything works

Issue #406, opened 2023-02-24
Missing NUL for null-terminated string


ccli8:
The issue can reproduce just by code review. Check the code fragment below:
https://github.com/Azure/iot-hub-device-update/blob/2d1f03671d45be1e55b89c940514ffe91b3227e0/src/utils/jws_utils/src/jws_utils.c#L214-L222

`tempHeader` should append with `NUL` to be a valid null-terminated string:
```C
strncpy(tempHeader, jws, headerLen);
tempHeader[headerLen] = '\0';
```

jw-msft:
will add unit test to cover this particular issue. Across the board will replace all instances of strn*xxx and replace with equivalent safer ones (possibly STRING_*). Also, looking for gcc compiler flag or static analysis to add to pipeline to prevent future occurences.

jw-msft:
I opted for a more targeted fix to start with.  Will address the rest of code base in separate PR(s).

jw-msft:
Thanks for the report, @ccli8. Resolved with this commit:
https://github.com/Azure/iot-hub-device-update/commit/e40b7e3d611d7fe9d6af99af325eb550ff0d5c49

PR #426, opened 2023-03-13
null terminate after strncpy of jws header (#417)

jw-msft:
- Cherry-Pick the following commit to main from develop branch: https://github.com/Azure/iot-hub-device-update/commit/e40b7e3d611d7fe9d6af99af325eb550ff0d5c49
- Originally reported by github issue https://github.com/Azure/iot-hub-device-update/issues/406

### DIFF
--- a/src/utils/jws_utils/src/jws_utils.c
+++ b/src/utils/jws_utils/src/jws_utils.c
@@ -219,6 +219,7 @@ static bool ExtractJWSHeader(const char* jws, char** header)
     }
 
     strncpy(tempHeader, jws, headerLen);
+    tempHeader[headerLen] = '\0';
     success = true;
 
 done:


src/utils/jws_utils/src/jws_utils.c
```diff
@@ -219,6 +219,7 @@ static bool ExtractJWSHeader(const char* jws, char** header)
     }
 
     strncpy(tempHeader, jws, headerLen);
```
JeffMill, line 221: `src\utils\c_utils\src\string_c_utils.c`

```C
/**
 * @brief Allocates memory for @p target and copy @p len characters of @p source into @p target buffer.
 *
 * @param[out] target Output string
 * @param source Source string
 * @param len Length of string to copy.
 * @return bool Returns true is success.
 */
bool MallocAndSubstr(char** target, char* source, size_t len)
```

Replace all malloc/strncpy with calls to this rather than doing 1 offs.

and remove this in that code:

```C
    if (strncpy(t, source, len) != t)
    {
        free(t);
        return false;
    }
```

strncpy **cant** fail.  It returns source.

Having said all that, I'm not quite sure MallocAndSubstr is null-terminating either!

JeffMill (changes requested):

src/utils/jws_utils/src/jws_utils.c
jw-msft: this and other uses of strncpy will be addressed by removing its usage across the code base in develop branch.  The intent of this PR is for cherry-picking the low-risk fix of this particular issue.

JeffMill (approved):

PR #431, opened 2023-03-18
Add Ubuntu Core snap update example document and artifacts 

Nox-MSFT:


yuvalg-MSFT (approved):
do we need one or more of our PM to view the docs?

Nox-MSFT:
> do we need one or more of our PM to view the docs?

Would be great, but feedback can come later, and we can incrementally improve this document.

PR #436, opened 2023-03-22
Reduce zlog buf BSS loader memory allocation from 512KB to 41KB

jw-msft:
# Summary 

This reduces the overall memory working set of the AducIotAgent by 471 KB.

When launching an ELF program, the linux loader reads the BSS metadata that was added by compiler and linker and allocates that amount of virtual memory upon execution.  After this change, the loader allocates only 41 KB for `_zlog_buffer` instead of 512KB as before.

## How were the new values determined?

I did the deployments as described below and ran the script at [this gist I created](https://gist.github.com/jw-msft/91c40fa84c396890a69aa357db894b92) on /var/log/adu directory to get the following stats.

Avg log line length was ~120 characters long and there was usually not more than 200 lines in the log files, although du-agent.* logs could be up to about 900 lines, so 128 max lines before a flush seemed like a reasonable default to reduce memory size by about an order of magnitude (10s of KB instead of 100s of KB).  There will be a few more flushes for a deployment for larger log files such as du-agent one, but that's probably ok especially when you want it to hit disk and have diagnostics interface upload it to azure blob storage.

Max line length that uses the `_zlog_buffer` (and does not write directly to file handle when exceeding max line length for things like the apply deployment C2D message) I saw was about 200 characters, so `328 (41 * 8)` seemed like a reasonable safety margin to avoid it flushing and writing directly to the file handle for almost all log messages that are traced save for a few really large ones (dumping serialized JSON message).

# mprof graphs of before and after

The following are sample of AducIotAgent memory usage every 0.1 sec using mprof (`pip install memory_profiler`).

Each graph shows the lifetime from very start of AducIotAgent binary and it downloading and deploying a 100MB update payload.  It's a bit difficult to see even after clicking the .png to zoom in as mprof does not produce the best looking PNG files.

## Description of baseline graph

The first baseline graph shows allocation just under 15 MB of virtual memory eventually going up a bit higher than 17.5 MB.

## Description of post-change graph

The 2nd graph for after this change shows it quit a bit more under 15 MB initially and it stays quite a bit below the 17 MB mark, so the graph of memory usage is essentially shifted lower, presumably by about 471 KB, so there is some sizeable memory savings with this change as would be expected.

### Baseline 512KB _zlog_buffer

mprof_baseline

### After Reduction to 41KB _zlog_buffer

mprof_AducIotAgent_reduced_zlog_buffer_BSS


src/logging/zlog/inc/zlog-config.h
```diff
@@ -15,8 +15,8 @@
 // enabling this will slow down the log
 // #define ZLOG_FORCE_FLUSH_BUFFER
 
-#define ZLOG_BUFFER_LINE_MAXCHARS 512
```
JeffMill, line 18: add comment as to how these values were derived so that someone doesnt come alone later and say "these are too small! I'm going to change MAXLINES to 1024"

JeffMill (changes requested):

JeffMill (approved):

PR #438, opened 2023-03-26
Add cmake fn to add test root keys only for debug builds

jw-msft:
- Add embed_test_root_keys_if_applicable cmake function into new security.cmake
- Invoke embed_test_root_keys_if_applicable instead of unconditionally including test root keys
- Ran cmake-format.sh
- Verified the following using `objdump -s out/src/utils/root_key_utils/CMakeFiles/root_key_utils.dir/src/root_key_lists.c.o`:
  - `--type=Debug`  has 2 retail *and 2 test keys*
  - `--type=RelWithDebInfo` has only 2 retail keys
  -   `--type=Relelease` has only 2 retail keys
  -   `--type=MinSizeRel` has only 2 retail keys

### DIFF
--- a/cmake/aduc_security.cmake
+++ b/cmake/aduc_security.cmake
@@ -1,0 +1,20 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+#
+# Embeds the test root keys into the target binary if applicable.
+#
+# Inputs: target - The cmake target that may be affected.
+#
+# Dependencies: CMAKE_BUILD_TYPE - The build type factors into applicability decision.
+# Side-effects: When applicable, the target's compile definitions will be affected.
+#
+function (embed_test_root_keys_if_applicable target)
+    message (STATUS "  CMAKE_BUILD_TYPE is: " ${CMAKE_BUILD_TYPE})
+    if ("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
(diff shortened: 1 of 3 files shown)


Nox-MSFT (approved):

nihemstr (approved):

cmake/aduc_security.cmake
```diff
@@ -0,0 +1,20 @@
...
+#
+# Embeds the test root keys into the target binary if applicable.
+#
+# Inputs: target - The cmake target that may be affected.
+#
+# Dependencies: CMAKE_BUILD_TYPE - The build type factors into applicability decision.
+# Side-effects: When applicable, the target's compile definitions will be affected.
+#
+function (embed_test_root_keys_if_applicable target)
+    message (STATUS "  CMAKE_BUILD_TYPE is: " ${CMAKE_BUILD_TYPE})
```
JeffMill, line 13: nit: this seems like unnecessary tracing.

cmake/aduc_security.cmake
```diff
@@ -0,0 +1,20 @@
...
+    if ("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
+        message (STATUS "Adding Test Root Keys for Debug Build.")
```
JeffMill, line 15: what's the implications of this? what happens if a customer releases a debug build (intentionally or not)? Are root keys embedded?

JeffMill (changes requested):

PR #442, opened 2023-03-27
Move extern Default_ExtensionManager_Download_Options into header

jw-msft:
[Bug 42998192](https://microsoft.visualstudio.com/OS/_workitems/edit/42998192): Why isn't Default_ExtensionManager_Download_Options extern in extension_manager.hpp ?

### DIFF
--- a/src/extensions/extension_manager/inc/aduc/extension_manager_download_options.h
+++ b/src/extensions/extension_manager/inc/aduc/extension_manager_download_options.h
@@ -1,6 +1,8 @@
 #ifndef ADUC_EXTENSION_MANAGER_DOWNLOAD_OPTIONS_H
 #define ADUC_EXTENSION_MANAGER_DOWNLOAD_OPTIONS_H
 
+#include <aduc/c_utils.h>
+
 // Default Content Downloader Download Timeout of 8 hours.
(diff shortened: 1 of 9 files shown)


src/extensions/extension_manager/inc/aduc/extension_manager_download_options.h
```diff
@@ -10,4 +12,8 @@ typedef struct tagADUC_ExtensionManager_Download_Options
         timeoutInMinutes; /**< The maximum number of minutes the content downloader should wait for the download to complete(whilst the network interface stays up). */
 } ExtensionManager_Download_Options;
 
+EXTERN_C_BEGIN
+extern ExtensionManager_Download_Options Default_ExtensionManager_Download_Options;
```
JeffMill, line 16: why is this marked as "extern"? Is the definition in external linkage (e.g. a separate module)?

JeffMill (approved):
approved with comment

src/extensions/extension_manager/inc/aduc/extension_manager_download_options.h
jw-msft: Yes, the definition is in extension_manager_helper.cpp.  It used to be in extension_manager.cpp, but that lead to a duplicate, conflicting definition because extension_manager.cpp includes extension_manager_download_options.h.

PR #444, opened 2023-03-28
Fix Bug: IsConnectionInfoValid linking to wrong methods?

shiyi-peng:
ADUC_ConnType GetConnTypeFromConnectionString(const char* connectionString);
bool GetConnectionInfoFromConnectionString(ADUC_ConnectionInfo* info, const char* connectionString);
bool GetConnectionInfoFromIdentityService(ADUC_ConnectionInfo* info);
bool GetAgentConfigInfo(ADUC_ConnectionInfo* info);
All functions have duplicates in main.c and iothub_communication_manager.c, fixing this.


src/communication_managers/iothub_communication_manager/inc/aduc/iothub_communication_manager.h
```diff
@@ -13,6 +13,40 @@
...
+ * @brief Scans the connection string and returns the connection type related to the string
+ * @details The connection string must use the valid, correct format for the DeviceId and/or the ModuleId
+ * e.g.
+ * "DeviceId=some-device-id;ModuleId=some-module-id;"
+ * If the connection string contains the DeviceId it is an ADUC_ConnType_Device
+ * If the connection string contains the DeviceId AND the ModuleId it is an ADUC_ConnType_Module
+ * @param connectionString the connection string to scan
+ * @returns the connection type for @p connectionString
+ */
+ADUC_ConnType GetConnTypeFromConnectionString(const char* connectionString);
```
JeffMill, line 26: i'm not 100% sure what belongs in "iothub_communication_manager", but none of these methods seem like they're related to iothub?

JeffMill (changes requested):

src/communication_managers/iothub_communication_manager/inc/aduc/iothub_communication_manager.h
jw-msft: agreed. Factoring out into a connectioninfo_utils static lib under src/utils seems reasonable to me.

src/communication_managers/iothub_communication_manager/inc/aduc/iothub_communication_manager.h
```diff
@@ -13,6 +13,40 @@
...
+
+/**
+ * @brief Get the Connection Info from connection string, if a connection string is provided in configuration file
+ *
+ * @return true if connection info can be obtained
+ */
+bool GetConnectionInfoFromConnectionString(ADUC_ConnectionInfo* info, const char* connectionString);
```
jw-msft, line 33: Are these used outside of iothub_communication_manager.c?  If not, there is no need to expose them externally (it can prevent linker from removing unused code)

src/communication_managers/iothub_communication_manager/src/iothub_communication_manager.c
```diff
@@ -440,7 +440,7 @@ static bool ADUC_DeviceClient_Create(
  * @param connectionString the connection string to scan
  * @returns the connection type for @p connectionString
  */
-static ADUC_ConnType GetConnTypeFromConnectionString(const char* connectionString)
```
jw-msft, line 443: If these are not used outside of iothub_communication_manager.c, then let's keep the static keywords so that they are not externally linked.

jw-msft (approved):
Sign off with comments with expectation that factoring out to util and add UTs will be done in another PR/commit.

src/communication_managers/iothub_communication_manager/inc/aduc/iothub_communication_manager.h
JeffMill: @shiyi-peng  please create follow up bug / task for proper refactoring + UT

shiyi-peng: @JeffMill , np , [Task 43982458](https://microsoft.visualstudio.com/OS/_workitems/edit/43982458): [P2] Refactoring iothub_communication_manager functions to a utils file and add UTs thank you!

JeffMill (approved):

PR #445, opened 2023-03-28
Fix Bug: Connection retry uses fuzzy math (perhaps just a tracing issue?)

shiyi-peng:
Bug: The `g_next_authentication_attempt_time` variable is updated in the `Connection_Maintenance` function, but the `IoTHub_CommunicationManager_ConnectionStatus_Callback` function may not always be in sync with the latest value of `g_next_authentication_attempt_time`. This can cause a discrepancy between the actual time until the next authentication attempt and the time reported in the log message.

### DIFF
--- a/src/communication_managers/iothub_communication_manager/src/iothub_communication_manager.c
+++ b/src/communication_managers/iothub_communication_manager/src/iothub_communication_manager.c
@@ -222,10 +222,11 @@ void IoTHub_CommunicationManager_ConnectionStatus_Callback(
         }
         else
         {
+            int time_until_next_retry = g_next_authentication_attempt_time - now_time;
             Log_Error(
                 "IoTHub connection is broken for %d seconds (will retry in %d seconds)",
                 now_time - g_first_unauthenticated_time,
-                g_next_authentication_attempt_time - now_time);
+                time_until_next_retry);
         }
         break;
     }


src/communication_managers/iothub_communication_manager/src/iothub_communication_manager.c
```diff
@@ -222,10 +222,11 @@ void IoTHub_CommunicationManager_ConnectionStatus_Callback(
         }
         else
         {
+            int time_until_next_retry = g_next_authentication_attempt_time - now_time;
             Log_Error(
                 "IoTHub connection is broken for %d seconds (will retry in %d seconds)",
                 now_time - g_first_unauthenticated_time,
-                g_next_authentication_attempt_time - now_time);
+                time_until_next_retry);
```
JeffMill, line 229: i can't figure out why moving the definition outside of Log_Error would fix this?

JeffMill (changes requested):

src/communication_managers/iothub_communication_manager/src/iothub_communication_manager.c
```diff
@@ -222,10 +222,11 @@ void IoTHub_CommunicationManager_ConnectionStatus_Callback(
         }
         else
         {
+            int time_until_next_retry = g_next_authentication_attempt_time - now_time;
```
jw-msft, line 225: The typedef for time_t is likely 64-bits on a modern-day system, and the int type here will cause an overflow as per the [Year 2038 problem](https://en.wikipedia.org/wiki/Year_2038_problem).  ISO C only guarantees time_t to be an arithmetic type--technically it can be floating point on some platforms.  POSIX-compliant platforms guarantee it is an integral type.  You should be able to use the latter (`long int`) and corresponding format specifier (`%ld`).  It is also a `long int` on windows platform if you search [here](https://learn.microsoft.com/en-us/cpp/c-runtime-library/standard-types?view=msvc-170).

PR #448, opened 2023-03-31
Added code to create the rootkey store when it does not exist

nihemstr:
- Added code to create the rootkey store when it does not exist in the rootkey_workflow
- Bug for code: https://microsoft.visualstudio.com/OS/_queries/edit/43774835/?triage=true


src/utils/rootkeypackage_utils/src/rootkeypackage_do_download.cpp
```diff
@@ -20,7 +20,7 @@ ADUC_Result DownloadRootKeyPkg_DO(const char* url, const char* targetFilePath)
     Log_Info("Downloading File '%s' to '%s'", url, targetFilePath);
 
     const std::error_code ret = microsoft::deliveryoptimization::download::download_url_to_path(url, targetFilePath);
-    if (ret == std::error_code())
+    if (! ret)
```
jw-msft, line 23: should it be doing ret.value() explicit == check with 0 ?

nihemstr: I was following the standard here:

``` const std::error_code doErrorCode = MSDO::download::download_url_to_path(
    entity->DownloadUri, fullFilePath.str(), false, std::chrono::seconds(retryTimeout));
    if (!doErrorCode)
    {
        resultCode = ADUC_Result_Download_Success;
        extendedResultCode = 0;
    }
```

I can explicitly check ret.value() == 0 if that is easier.

src/inc/aduc/result.h
```diff
@@ -79,117 +79,90 @@ MAKE_ADUC_EXTENDEDRESULTCODE(const unsigned int facility, const unsigned int com
  */
```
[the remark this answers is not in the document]
jw-msft: This is orthogonal to this PR. There is some value in result.h lookup value for now and can consider doxygen later.

jw-msft (approved):

PR #467, opened 2023-04-28
Register X-Compiled Arm64 Components before Zip

jw-msft:


scripts/install-du.ps1
```diff
@@ -7,6 +7,7 @@ Assumes the product has been built locally using build.ps1
 
 .EXAMPLE
 PS> .\scripts\du_install.ps1 -Type Debug -Install -Package
+PS> .\scripts\du_install.ps1 -Type Debug -TargetArch arm64 -Package -BuildOutputPath c:\src\iot-hub-device-update\out\ARM64-Debug
```
JeffMill, line 10: I'm not seeing `BuildOutputPath` as a Param.

scripts/install-du.ps1
```diff
@@ -16,7 +17,9 @@ Param(
     [ValidateSet('Release', 'RelWithDebInfo', 'Debug')]
     [string]$Type = 'Debug',
     [ValidateSet('x64', 'arm64')]
-    [string]$Arch = 'x64',
+    [string]$HostArch = 'x64',
```
JeffMill, line 20: I can understand renaming Arch to TargetArch (it's clearer) but what is "HostArch"? isn't obvious what system this script is currently running on?

scripts/install-du.ps1
```diff
@@ -94,30 +97,71 @@ function CalcHash {
     [Convert]::ToBase64String($hashAlg.ComputeHash([System.IO.File]::ReadAllBytes($FilePath)))
 }
```
JeffMill, line 98: nit: [System.] is implied, so you can just use [Security....] or [IO.File]

scripts/install-du.ps1
```diff
@@ -94,30 +97,71 @@ function CalcHash {
...
         hashes = @{
             sha256="$fileHash"
         }
     }
 
-    $registration | ConvertTo-Json | Set-Content $OutputJsonFilePath
+    $registration | ConvertTo-Json | Set-Content "$ExtensionsDir/content_downloader/extensions.json"
+}
+
+function WriteContentHandlerRegistration{
```
JeffMill, line 123: I originally didn't include this as I'm concerned about the format of the registration file changing and this breaking.  

Maybe a better approach would be rather than calling into the handler to do the registration as we do now, the handler has something like "--GetRegistrationJson --InstallationPath foo" which would output the json so that this script and the agent could use that to write the file.

scripts/install-du.ps1
```diff
@@ -94,30 +97,71 @@ function CalcHash {
...
+        [Parameter(Mandatory = $true)][string]$ExtensionsDir
+    )
+
+    $adu_extensions_sources_dir = "$ExtensionsDir/sources"
+
+    $binaryFilePath = "$adu_extensions_sources_dir/${BinaryFileNameNoExt}.dll"
+    $fileSizeInBytes = (Get-Item $binaryFilePath).Length
+    $fileHash = CalcHash -FilePath $binaryFilePath
+
+    $numUnderscore = ($ExtensionName | Select-String -Pattern '_' -AllMatches).Matches.Count
```
JeffMill, line 136: not clear to me why this is needed?

JeffMill: in any case, as I mention above, we shouldn't be creating our own registration code -- let's change the registration from self-registration to (or in addition?) outputting json

scripts/install-du.ps1
```diff
@@ -94,30 +97,71 @@ function CalcHash {
...
+    $registration = [ordered]@{
+        fileName="$binaryFilePath"
+        sizeInBytes=$fileSizeInBytes
+        hashes = @{
+            sha256="$fileHash"
+        }
+        handlerId = "$handlerId"
+    }
+
+    $registration | ConvertTo-Json | Set-Content "$ExtensionsDir/update_content_handlers/$ExtensionName/content_handler.json"
```
JeffMill, line 157: Set-Content -> Out-File ? (unclear what the diff is)

scripts/install-du.ps1
```diff
@@ -131,39 +175,28 @@ function Register-Components {
...
-    if ($TargetArch -eq "arm64") {
-        $adu_extensions_update_content_handlers_dir = "$adu_extensions_dir/update_content_handlers"
-
-        WriteExtensionRegistration -BinaryFilePath $do_content_downloader_file `
-            -OutputJsonFilePath "$adu_extensions_dir/content_downloader/extension.json"
 
-        $chName = 'content_handler.json'
-        WriteExtensionRegistration -BinaryFilePath $microsoft_script_1_handler_file `
-            -OutputJsonFilePath "$adu_extensions_update_content_handlers_dir/microsoft_script_1/$chName"
+    # Cross-compiled target means we cannot use agent binary to create the registrations, so create them in script.
```
JeffMill, line 179: interesting -- my suggestion above (have the handler output json wont work on cross-compilation as we wont be able to run the binaries natively.

Maybe there's some alternative way? (have a registration template with replacable parmaeters that agnet and this code uses?)

JeffMill: I wasn't thrilled about having to register components locally in order to package them, so if we can fix this up that would be great.

scripts/install-du.ps1
JeffMill: i looked below -- it would be preferable not to do any registration in this file at all, which should remove the need for a hostarch (or analyzing system)

JeffMill (changes requested):

scripts/install-du.ps1
jw-msft: It's because the handlerId for "update manifest" handler only has one underscore, that is, no `"_1"` version at the end: e.g. `"microsoft_update-manifest"` which will get converted to `"microsoft/update-manifest"`

The handlerId of "step handlers" always have 2 underscores because they are always versioned:
`"microsoft_apt_1" -> "microsoft/apt:1`

scripts/install-du.ps1
jw-msft: Found this:
https://stackoverflow.com/questions/10655788/powershell-set-content-and-out-file-what-is-the-difference

Yes, switching to Out-File. Set-Content can do some things that Out-File can't but unneeded here.

scripts/install-du.ps1
jw-msft: yes, I was trying to avoid analyzing system arch for short-term, but I think I like your registration template approach so that we don't have to do any logic other than search and replace.

scripts/install-du.ps1
jw-msft: I like registration template approach. Will work on that.

jw-msft:
Closing to replace with new approach of installer script in the zip, or just switchover to .msi

PR #469, opened 2023-05-03
Rework the end-to-end test scenario definitions/runners

shiyi-peng:
* Before the change,  we have scenarioPath (e.g. scenarios/ubuntu-20.04-arm64/ ) that have all test cases for each distro+ arch. Now I consolidate all scenarios & all test cases to `scenarios/test_runner`. 
* In file scenario_definitions.py, I define `DuScenarioDefinitionManage` that uses parameter `distroName` to determine to behavior for the responding scenario.


azurepipelines/e2e_test/scenarios/test_runner/vm_setup/setup.sh
```diff
@@ -177,11 +255,10 @@ function test_shutdown_service() {
     done
 }
 
-configure_apt_repository
+configure_apt_repository "$distro" "$architecture"
```
nihemstr, line 258: IS this on the latest commit from develop? It looks like it's missing my AIS SAS work (same as the above runners).

azurepipelines/e2e_test/scenarios/test_runner/Bundle-update.py
```diff
@@ -49,6 +50,16 @@ def test_AptDeployment(self):
...
+        self.aduScenarioDefinition = DuScenarioDefinitionManager.FromOSEnvironment()
+
+        test_device_id = self.aduScenarioDefinition.test_device_id
+        test_adu_group = self.aduScenarioDefinition.test_adu_group
+        global test_result_file_prefix
+        test_result_file_prefix = self.aduScenarioDefinition.test_result_file_prefix
+        test_bundle_update_deployment_id = self.aduScenarioDefinition.test_bundle_update_deployment_id
+        test_connection_timeout_tries = self.aduScenarioDefinition.test_connection_timeout_tries
+        retry_wait_time_in_seconds = self.aduScenarioDefinition.retry_wait_time_in_seconds
+
```
nihemstr, line 62: The below commits don't look like they track from the latest merge to develop (completed Monday). Is there any chance we're missing those commits?

azurepipelines/e2e_test/scenarios/test_runner/apt_deployment.py
```diff
@@ -48,6 +49,16 @@ def test_AptDeployment(self):
         self.aduTestConfig = DuAutomatedTestConfigurationManager.FromOSEnvironment()
         self.duTestHelper = self.aduTestConfig.CreateDeviceUpdateTestHelper()
 
+        self.aduScenarioDefinition = DuScenarioDefinitionManager.FromOSEnvironment()
+
+        test_device_id = self.aduScenarioDefinition.test_device_id
+        test_adu_group = self.aduScenarioDefinition.test_adu_group
+        global test_result_file_prefix
```
nihemstr, line 56: Globals scare me. What's the use case for declaring it here within all the tests?

azurepipelines/e2e_test/scenarios/test_runner/scenario_definitions.py
```diff
@@ -0,0 +1,52 @@
...
+from testingtoolkit import UpdateId
+
+#
+# Base Device Definitions
+#
+
+class DuScenarioDefinitionManager:
+    def __init__(self):
+        self.test_device_id = ""
+        self.test_adu_group = ""
```
nihemstr, line 24: It's better to have these as getter/setter model instead of using direct access in the tests so we can prevent any mistakes when accessing them in the individual tests

nihemstr (dismissed):
Approved with some comments.

azurepipelines/e2e_test/e2etest.yaml
```diff
@@ -130,7 +130,7 @@ stages:
                       du_tarball_script: >-
                           tar -xvf /tmp/testsetup.tar.gz -C ./ &&
                           chmod u=rwx,g=rwx,o=rx ./testsetup/setup.sh &&
-                          ./testsetup/setup.sh --self-upgrade
+                          ./testsetup/setup.sh --distro ubuntu-18.04 --architecture amd64
```
Nox-MSFT, line 133: what happened to '--self-upgrade' option?

azurepipelines/e2e_test/scenarios/test_runner/vm_setup/setup.sh
```diff
@@ -23,45 +23,90 @@
...
+    echo "Error: Both distro and architecture must be provided"
+    print_help
+    exit 1
+fi
+
 #
-# Install the Microsoft APT repository
+# Install Device Update Dependencies from APT
 #
+# Note: If there are other dependencies tht need to be installed via APT or other means they should
```
Nox-MSFT, line 74: typo: "tht"

Nox-MSFT (changes requested):

azurepipelines/e2e_test/e2etest.yaml
shiyi-peng: I have a separate PR that add supports self upgrade , removing it here for the simplicity of this PR

Nox-MSFT (dismissed):

azurepipelines/e2e_test/scenarios/test_runner/apt_deployment.py
shiyi-peng: I am not familiar with how exactly this works, it uses test_result_file_prefix outside the class. I am not sure what's the best way to do it?:
if (__name__ == "__main__"):
    #
    # Create the IO pipe
    #
    out = io.BytesIO()

    #
    # Exercise the TestCase and all the tests within it.
    #
    unittest.main(testRunner=xmlrunner.XMLTestRunner(output=out),
                  failfast=False, buffer=False, catchbreak=False, exit=False)

    #
    # Finally transform the output unto the X/JUnit XML file format
    #
    with open('./testresults/' + test_result_file_prefix + '-apt-deployment-test.xml', 'wb') as report:
        report.write(transform(out.getvalue()))

nihemstr (approved):

PR #472, opened 2023-05-04
Add 'apiVersion' support for swupdate handler v2 and script handler

Nox-MSFT:
## Description

- Simplify the arguments and options the SWUpdate Handler and Script Handler send to the scripts. 
To ensure compatibility with previous versions, the feature will be activated when the handlerProperties["apiVersion"] is set to "2.1".  
In case the handlerProperties["apiVersion"] is not specified, the handler will revert to the previous behavior.
The previous behavior involved the handler passing arguments like "--action-is-installed", "--action-download", "--action-install", and so on.
Under the new behavior (apiVersion 2.1), the handler will pass arguments such as "--action is-installed", "--action download", "--action install", and so on.

- Also, added Script Handler unit tests 
- Update swupdate handler unit tests to reflect apiVersion option

**Internal tracking number : 43101619**


src/extensions/step_handlers/script_handler/src/script_handler.cpp
```diff
@@ -22,6 +22,8 @@
 #include <vector>
 
 #define HANDLER_PROPERTIES_SCRIPT_FILENAME "scriptFileName"
+#define HANDLER_PROPERTIES_API_VERSION "apiVersion"
+#define HANDLER_ARG_ACTION "--action"
```
JeffMill, line 26: these are breaking changes, and it looks like you've added an apiVersion, but what does this mean for someone who has already written a script (*) -- the new agent would break their script.

(*) there probably aren't any, so it's likely okay for now?

src/extensions/step_handlers/script_handler/src/script_handler.cpp
```diff
@@ -236,7 +238,7 @@ ADUC_Result ScriptHandlerImpl::Download(const tagADUC_WorkflowData* workflowData
     }
 
     // Invoke primary script to download additional files, if required.
-    result = PerformAction("--action-download", workflowData);
+    result = PerformAction("download", workflowData);
```
JeffMill, line 241: nit: would it be better to pass an enum value, e.g. { Download, Install, ... } rather than a string, in case these values change, or if PerformAction needs to do actions based on the action, e.g. "if action == Download" is better tahn "if (strcmp(action, "Download"))"

src/extensions/step_handlers/script_handler/src/script_handler.cpp
```diff
@@ -501,19 +503,38 @@ ADUC_Result ScriptHandlerImpl::PrepareScriptArguments(
...
+ * @param[in] args List of options and arguments.
+ * @param[out] commandLineArgs An output command-line arguments.
+ * @param[out] scriptOutput If @p prepareArgsOnly is false, this will contains the action output string.
+ * @return ADUC_Result
+ */
+ADUC_Result ScriptHandler_PerformAction(
+    const std::string& action,
+    const tagADUC_WorkflowData* workflowData,
+    bool prepareArgsOnly,
+    std::string& scriptFilePath,
```
JeffMill, line 528: bunch of non-const references?  are these out parameters?  If so, this would be much better as a struct/class that's returned, e.g.

ActionValues {
  string args;
  string output;
  ADUC_Result result;
}

ActionValues PerformAction()

src/extensions/step_handlers/script_handler/src/script_handler.cpp
```diff
@@ -552,14 +578,32 @@ static ADUC_Result ScriptHandler_PerformAction(const std::string& action, const
...
-    aduShellArgs.emplace_back(action.c_str());
+    // Prepare arguments based on specified api version.
+    if (apiVer == nullptr || strcmp(apiVer, "2.0") == 0)
+    {
+        std::string backcompatAction = "--action-" + action;
+        aduShellArgs.emplace_back(adushconst::target_options_opt);
+        aduShellArgs.emplace_back(backcompatAction.c_str());
+        commandLineArgs.emplace_back(backcompatAction.c_str());
+    }
+    else if (strcmp(apiVer, "2.1") == 0)
```
JeffMill, line 591: careful - string compares on versions are problematic.

If the user puts "2.01" is that "2.1"?  What about "2.10" is that "2.1"?  I think an integer would be better.

src/extensions/step_handlers/script_handler/src/script_handler.cpp
```diff
@@ -533,6 +557,8 @@ static ADUC_Result ScriptHandler_PerformAction(const std::string& action, const
                                               adushconst::update_action_opt,
                                               adushconst::update_action_execute };
 
+    std::stringstream ss;
```
JeffMill, line 560: move closer to usage

src/extensions/step_handlers/script_handler/src/script_handler.cpp
```diff
@@ -572,6 +616,26 @@ static ADUC_Result ScriptHandler_PerformAction(const std::string& action, const
         Log_Debug("##########\n# ADU-SHELL ARGS:\n##########\n %s", ss.str().c_str());
     }
 
+    if (prepareArgsOnly)
+    {
+        for (const auto& a : aduShellArgs)
+        {
+            if (a[0] != '-')
+            {
+                ss << " \"" << a << "\"";
```
JeffMill, line 625: what if 'a' contains a quote?

src/extensions/step_handlers/swupdate_handler_v2/tests/testdata/adu-yocto-ab-rootfs-update/payloads/example-a-b-update.sh
```diff
@@ -205,12 +205,17 @@ print_help() {
...
+    echo "                                              is-installed, download, install, apply,"
+    echo "                                              cancel, backup, restore."
+    echo ""
     echo "--action-is-installed                     Perform 'is-installed' check."
     echo "                                          Check whether the selected component [or primary device] current states"
     echo "                                          satisfies specified 'installedCriteria' data."
     echo "--installed-criteria                      Specify the Installed-Criteria string."
     echo ""
-    echo "--action-download                         Perform 'download' aciton."
+    echo "--action-download                         Perform 'download' action."
```
JeffMill, line 218: why include the "legacy" options (--action-*)

src/extensions/step_handlers/swupdate_handler_v2/tests/testdata/swupdate_filecopy/example-du-swupdate-script.sh
```diff
@@ -187,12 +187,17 @@ print_help() {
...
+    echo "                                              is-installed, download, install, apply,"
+    echo "                                              cancel, backup, restore."
+    echo ""
     echo "--action-is-installed                     Perform 'is-installed' check."
     echo "                                          Check whether the selected component [or primary device] current states"
     echo "                                          satisfies specified 'installedCriteria' data."
     echo "--installed-criteria                      Specify the Installed-Criteria string."
     echo ""
-    echo "--action-download                         Perform 'download' aciton."
+    echo "--action-download                         Perform 'download' action."
```
JeffMill, line 200: as above.

JeffMill (changes requested):

src/extensions/step_handlers/swupdate_handler_v2/tests/testdata/swupdate_filecopy/example-du-swupdate-script.sh
Nox-MSFT: For backward compatibility.

src/extensions/step_handlers/script_handler/src/script_handler.cpp
Nox-MSFT: I think this is ok. It's quite common to use "major.minor" without any leading zero.
For next iteration, if there's a breaking change, we could release another handler sxs that associates with the new version of update handler type. e.g. "microsoft/scripthandler:**x**"

src/extensions/step_handlers/script_handler/src/script_handler.cpp
Nox-MSFT: Not breaking changes. Just adding new features. Customer will get old behavior by default.

src/extensions/step_handlers/script_handler/src/script_handler.cpp
Nox-MSFT: Changed. Thanks for the suggestion.

src/extensions/step_handlers/script_handler/src/script_handler.cpp
Nox-MSFT: Done.

src/extensions/step_handlers/script_handler/src/script_handler.cpp
Nox-MSFT: Added code to quote using ' for the case that the string contains "

jw-msft:
It could be opportune time to address https://github.com/Azure/iot-hub-device-update/discussions/488 by putting info from workflowData and workflowHandle such as cancellation type (as per discussion) into adu-shell environment var context so script author can just do $FOO instead of parsing script cmdline args.

src/extensions/step_handlers/script_handler/src/script_handler.cpp
```diff
@@ -544,22 +560,40 @@ static ADUC_Result ScriptHandler_PerformAction(const std::string& action, const
...
+    {
+        aduShellArgs.emplace_back(adushconst::target_options_opt);
+        aduShellArgs.emplace_back(HANDLER_ARG_ACTION);
+        results.commandLineArgs.emplace_back(HANDLER_ARG_ACTION);
 
-    aduShellArgs.emplace_back(adushconst::target_options_opt);
-    aduShellArgs.emplace_back(action.c_str());
+        aduShellArgs.emplace_back(adushconst::target_options_opt);
+        aduShellArgs.emplace_back(action.c_str());
+        results.commandLineArgs.emplace_back(action.c_str());
```
jw-msft, line 589: maybe outside scope of this PR but it'd be really nice to export all the work data and workflow handle data that could be useful as env var for the scripts to consume without having to parse cmdline opts

Nox-MSFT: Will do this as part of  Internal Task#43162147]

src/extensions/step_handlers/script_handler/src/script_handler.cpp
```diff
@@ -501,30 +504,43 @@ ADUC_Result ScriptHandlerImpl::PrepareScriptArguments(
...
     {
         Log_Error("Workflow data or handler is null. This is unexpected!");
-        result.ExtendedResultCode = ADUC_ERC_SCRIPT_HANDLER_INSTALL_ERROR_NULL_WORKFLOW;
-        return result;
+        results.result.ExtendedResultCode = ADUC_ERC_SCRIPT_HANDLER_INSTALL_ERROR_NULL_WORKFLOW;
+        return results;
     }
 
+    const char* apiVer = workflow_peek_update_manifest_handler_properties_string(
+        workflowData->WorkflowHandle, HANDLER_PROPERTIES_API_VERSION);
```
jw-msft, line 539: I don't think we should allow specifying contract version in the handler properties of update metadata, but it should be tied to the version of the apt handler .so that is installed instead.  Reasoning is ALL past versions of every content handler would need to be installed/included with the .deb pkg. Instead, I think only installing the latest version makes sense.

src/extensions/step_handlers/script_handler/src/script_handler.cpp
```diff
@@ -544,22 +560,40 @@ static ADUC_Result ScriptHandler_PerformAction(const std::string& action, const
...
         goto done;
     }
 
     aduShellArgs.emplace_back(adushconst::target_data_opt);
-    aduShellArgs.emplace_back(scriptFilePath);
+    aduShellArgs.emplace_back(results.scriptFilePath);
+    results.commandLineArgs.emplace_back(results.scriptFilePath);
+
+    // Prepare arguments based on specified api version.
+    if (apiVer == nullptr || strcmp(apiVer, "2.0") == 0)
```
jw-msft, line 574: It should just use contract version pattern that is used for all the other extensions instead of diverging with new way. The struct has major and minor integer fields in the struct.  https://github.com/Azure/iot-hub-device-update/blob/develop/src/utils/contract_utils/inc/aduc/contract_utils.h#L23

src/extensions/step_handlers/swupdate_handler_v2/src/swupdate_handler_v2.cpp
```diff
@@ -185,9 +188,24 @@ ADUC_Result SWUpdateHandler_PerformAction(
...
-    commandLineArgs.emplace_back(action.c_str());
+    // Prepare arguments based on specified api version.
+    if (apiVer == nullptr || strcmp(apiVer, "2.0") == 0)
+    {
+        std::string backcompatAction = "--action-" + action;
+        aduShellArgs.emplace_back(adushconst::target_options_opt);
+        aduShellArgs.emplace_back(backcompatAction.c_str());
+        commandLineArgs.emplace_back(backcompatAction.c_str());
+    }
+    else if (strcmp(apiVer, "2.1") == 0)
```
jw-msft, line 199: Any reason why we could not use ADUC_ExtensionContractInfo? It is used for shared lib exports but could also be used for this shell-scripting contract.

src/extensions/step_handlers/swupdate_handler_v2/src/swupdate_handler_v2.cpp
```diff
@@ -185,9 +188,24 @@ ADUC_Result SWUpdateHandler_PerformAction(
     aduShellArgs.emplace_back(scriptFilePath);
     commandLineArgs.emplace_back(scriptFilePath);
 
-    aduShellArgs.emplace_back(adushconst::target_options_opt);
-    aduShellArgs.emplace_back(action.c_str());
-    commandLineArgs.emplace_back(action.c_str());
+    // Prepare arguments based on specified api version.
+    if (apiVer == nullptr || strcmp(apiVer, "2.0") == 0)
```
jw-msft, line 192: Maybe it needs more offline discussion, but I don't think we should have code for older shell script api. There is contract version that is tied with handler version (e.g. microsoft/swupdate:2), so it seems reasonable to only support latest "shell script arg passing" contract/apiVer; otherwise, test matrix is multiplied.  Thinking is it should be ok as long as captured in release notes as a breaking change so that script authoring can be changed (will be tested before deploying new agent to production devices).

jw-msft (changes requested):
commented on support story of older api ver and leveraging ADUC_ExtensionContractInfo for apiVer. There are 3 sets of versions: handler version has one or more contract versions (function export call sigs and semantics), and now shell-script calling convention version (i.e. apiVer). Need to understand back-compat story and if it makes sense to only support latest of each version set and document as breaking change.

src/extensions/step_handlers/script_handler/src/script_handler.cpp
jw-msft, line 574: why starting at "2.0". Would "1.0" work?

src/extensions/step_handlers/swupdate_handler_v2/src/swupdate_handler_v2.cpp
Nox-MSFT: Based on our previous offline discussions, this PR represents a significant enhancement for the libmicrosoft_script_handler v1. Customers now have the option to utilize the new features by including "apiVer":"2.1" in the import manifest. If the apiVer is not specified, customers will continue to use the script handler v1 as they did before. This update provides increased flexibility and improved functionality for our customers.

src/extensions/step_handlers/script_handler/src/script_handler.cpp
```diff
@@ -572,56 +606,92 @@ static ADUC_Result ScriptHandler_PerformAction(const std::string& action, const
...
+        std::stringstream ss;
+        for (const auto& a : aduShellArgs)
+        {
+            char quote = '\"';
+            char whitespace = ' ';
+
+            // NOLINTNEXTLINE(cppcoreguidelines-avoid-magic-numbers,readability-magic-numbers)
+            if (a[0] != '-')
+            {
+                // If the argument contains a quote character, then use the other quote character.
```
jw-msft, line 620: what if there's a mix of single and double quotes, or escaped quotes?

src/extensions/step_handlers/script_handler/src/script_handler.cpp
```diff
@@ -572,56 +606,92 @@ static ADUC_Result ScriptHandler_PerformAction(const std::string& action, const
...
-        Log_Info(scriptOutput.c_str());
+        Log_Info(results.scriptOutput.c_str());
     }
 
     if (exitCode != 0)
     {
         int extendedCode = ADUC_ERC_SCRIPT_HANDLER_CHILD_PROCESS_FAILURE_EXITCODE(exitCode);
         Log_Error("Script failed (%s), extendedResultCode:0x%X (exitCode:%d)", action.c_str(), extendedCode, exitCode);
-        result = { .ResultCode = ADUC_Result_Failure, .ExtendedResultCode = extendedCode };
+        results.result = { .ResultCode = ADUC_Result_Failure, .ExtendedResultCode = extendedCode };
```
jw-msft, line 653: use `/* .ResultCode = */ ADUC_ResultFailures` instead.
MSVC compiler requires C++20 for this syntax, which is c99 syntax actually.

jw-msft (commented):
approve with minor comment about starting with 1.0 apiVer

src/extensions/step_handlers/script_handler/src/script_handler.cpp
```diff
@@ -572,56 +606,92 @@ static ADUC_Result ScriptHandler_PerformAction(const std::string& action, const
...
         goto done;
     }
 
     // Parse result file.
     actionResultValue = json_parse_file(scriptResultFile.c_str());
     if (actionResultValue == nullptr)
     {
-        result = { .ResultCode = ADUC_Result_Failure,
-                   .ExtendedResultCode = ADUC_ERC_SCRIPT_HANDLER_INSTALL_FAILURE_PARSE_RESULT_FILE };
+        results.result = { .ResultCode = ADUC_Result_Failure,
```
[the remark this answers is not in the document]
Nox-MSFT: fixed

jw-msft (approved):

src/extensions/step_handlers/script_handler/src/script_handler.cpp
```diff
@@ -544,22 +560,40 @@ static ADUC_Result ScriptHandler_PerformAction(const std::string& action, const
...
         goto done;
     }
 
     aduShellArgs.emplace_back(adushconst::target_data_opt);
-    aduShellArgs.emplace_back(scriptFilePath);
+    aduShellArgs.emplace_back(results.scriptFilePath);
+    results.commandLineArgs.emplace_back(results.scriptFilePath);
+
+    // Prepare arguments based on specified api version.
+    if (apiVer == nullptr || strcmp(apiVer, "1.0") == 0)
```
JeffMill, line 574: still not thrilled with using string comparison for versions.  can this not be an integer?

src/extensions/step_handlers/script_handler/inc/aduc/script_handler.hpp
```diff
@@ -55,4 +69,15 @@ class ScriptHandlerImpl : public ContentHandler
...
+/**
+ * @brief Perform the specified action.
+ *
+ * @param action Name of the action to perform
+ * @param workflowData The workflow data
+ * @param prepareArgsOnly if true, only prepare the arguments and return them in the result
+ * @return ADUC_PerformAction_Results The result of the action and outputs of the script.
+ */
+ADUC_PerformAction_Results
+ScriptHandler_PerformAction(const std::string& action, const tagADUC_WorkflowData* workflowData, bool prepareArgsOnly);
```
JeffMill, line 81: not thrilled with passing action as a string -- wouldn't this be better as an enum value?

src/extensions/step_handlers/script_handler/src/script_handler.cpp
JeffMill: yeah, this still doesn't seem robust.

JeffMill (approved):
signed off with comments.

PR #474, opened 2023-05-09
Add Unit tests to ensure proper handling of downloader result.

jw-msft:
- Fix issue with overwriting DO err code with success in ExtensionManager::Download
- Add dep injection of resolving DownloadProc and add unit tests to ensure err is preserved
- Extract FileEntityWrapper RAII class to entity_utils as auto_entity_wrapper.hpp


Nox-MSFT (dismissed):

src/extensions/extension_manager/src/extension_manager.cpp
```diff
@@ -926,6 +938,10 @@ ADUC_Result ExtensionManager::Download(
         unsigned int timeoutInSeconds = 60 * timeoutInMinutes;
 
         result = downloadProc(entity, workflowId, workFolder.get(), timeoutInSeconds, downloadProgressCallback);
+        if (IsAducResultCodeFailure(result.ResultCode))
```
JeffMill, line 941: can you make a PR that only includes this fix and then another with all of your excellent additions (UTs, etc.)

JeffMill (changes requested):

src/extensions/extension_manager/src/extension_manager.cpp
```diff
@@ -945,6 +961,11 @@ ADUC_Result ExtensionManager::Download(
             goto done;
         }
     }
+    else
```
JeffMill, line 964: how does line 970+ ever get executed with this here?

src/extensions/extension_manager/src/extension_manager.cpp
```diff
@@ -807,11 +812,18 @@ ADUC_Result ExtensionManager::InitializeContentDownloader(const char* initialize
     return result;
 }
 
+DownloadProc ExtensionManager::DefaultDownloadProcResolver(void* lib, const char* symbol)
```
JeffMill, line 815: why pass in symbol?  There should only be one export that returns a DownloadProc, no?

JeffMill (changes requested):

src/extensions/extension_manager/src/extension_manager.cpp
jw-msft: removed

src/utils/entity_utils/inc/aduc/auto_file_entity.hpp
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * @brief Get the address of the file entity object.
+     * Useful for passing to methods with the signature "f(..., ADUC_FileEntity* outFileEntity)"
+     *
+     * @return ADUC_FileEntity* Non-const address of wrapped object.
+     * @details Will uninitialize the current field as the assumption is the entity will be reinitialized by the caller.
+     */
+    ADUC_FileEntity* address_of()
+    {
+        uninit();
```
Nox-MSFT, line 56: what don't we just make uninit() public, and require caller to explicitly call uninit() before calling address_of() ?

Nox-MSFT (dismissed):
Left 1 minor comment.

src/utils/entity_utils/inc/aduc/auto_file_entity.hpp
jw-msft: I see what you mean and it would make sense for const ptr.

I was aligning with similar pattern in STRING_HANDLE_wrapper for address_of() method: https://github.com/Azure/iot-hub-device-update/blob/feature/wiot-s1/src/utils/string_utils/inc/aduc/string_handle_wrapper.hpp#L56

I think if need a const pointer then can create a method for that later when needed; otherwise, the usage pattern is expecting that it will get overwritten so need to uninit.

jw-msft: fixed with simplification of this file

JeffMill (approved):

PR #477, opened 2023-05-16
winiot - Check volume label before format drive

jw-msft:
[Bug 44356009](https://microsoft.visualstudio.com/OS/_workitems/edit/44356009): [WinIoT] WIM ContentHandler needs to check Volume Label before formatting drive

### DIFF
--- a/src/extensions/step_handlers/wim_handler/src/format_drive.cpp
+++ b/src/extensions/step_handlers/wim_handler/src/format_drive.cpp
@@ -5,7 +5,6 @@
 #include <windows.h>
 
 #include <cstdio>
-
 #include <WbemCli.h>
 
 #define CHR(cmd)             \
@@ -17,7 +16,7 @@
         }                    \
     }
 
-HRESULT FormatDrive(char driveLetter)
+HRESULT FormatDrive(char driveLetter, const std::string& driveLabel)
 {
     CComPtr<IWbemLocator> spLocator;
     CHR(CoCreateInstance(CLSID_WbemLocator, NULL, CLSCTX_INPROC_SERVER, IID_PPV_ARGS(&spLocator)));
@@ -84,7 +83,7 @@ HRESULT FormatDrive(char driveLetter)
     CHR(spClassInstance->Put(L"FileSystem", 0, &varFileSystem, 0));
     CComVariant varQuickFormat(true);
     CHR(spClassInstance->Put(L"QuickFormat", 0, &varQuickFormat, 0));
-    CComVariant varLabel(L"IOT");
+    CComVariant varLabel(driveLabel.c_str());
     CHR(spClassInstance->Put(L"Label", 0, &varLabel, 0));
 
     CComPtr<IWbemClassObject> spOutParams;
@@ -98,7 +97,7 @@ HRESULT FormatDrive(char driveLetter)
     int formatResult = varReturnValue.intVal();
     if (formatResult != 0)
     {
-        return E_FAIL;
+        return HRESULT_FROM_WIN32(formatResult);
     }
 
     return S_OK;
--- a/src/extensions/step_handlers/wim_handler/src/wim_step_handler.cpp
+++ b/src/extensions/step_handlers/wim_handler/src/wim_step_handler.cpp
@@ -14,7 +14,6 @@
 #include <string>
 
 #include "aduc/logging.h"
-#include "aducpal/unistd.h" // ADUCPAL_access
 
 // TODO: This should be set in the update metadata.
 static const char TARGET_DRIVE = 'd';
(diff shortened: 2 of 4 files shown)


nihemstr (approved):

src/extensions/step_handlers/wim_handler/src/format_drive.cpp
```diff
@@ -17,8 +18,36 @@
         }                    \
     }
 
-HRESULT FormatDrive(char driveLetter)
+static HRESULT DoesVolumeLabelExist(char driveLetter, const std::string& volumeLabel)
```
JeffMill, line 21: i think this would be more useful as:

`std::string GetVolumelabel(char driveLetter);`

JeffMill: and have the caller do the compare.

src/extensions/step_handlers/wim_handler/src/format_drive.cpp
```diff
@@ -17,8 +18,36 @@
...
+    {
+        return HRESULT_FROM_WIN32(GetLastError());
+    }
+
+    return S_OK;
+}
+
+HRESULT FormatDrive(char driveLetter, const std::string& volumeLabel)
+{
+    CHR(DoesVolumeLabelExist(driveLetter, volumeLabel));
```
JeffMill, line 49: this doesn't belong here.  it belongs in the caller, as if I call FormatDrive I expect it to format the drive.

JeffMill (changes requested):

src/extensions/step_handlers/wim_handler/src/wim_step_handler.cpp
```diff
@@ -39,6 +41,76 @@ static std::string make_preferred(const char* path, const char* file = nullptr)
...
 
+/**
+ * @brief Verifies that the file path exists.
+ *
+ * @param payloadFilePath The path to the file.
+ * @return HRESULT S_OK when the file exists; an error hresult, otherwise.
+ */
+static HRESULT EnsurePayloadExists(const std::string& payloadFilePath)
+{
+    if (ADUCPAL_access(payloadFilePath.c_str(), F_OK) != 0)
```
JeffMill, line 52: this is win32-only code. we shouldn't be calling anything in ADUCPAL.

I'd suggest this:

```C++
BOOL FileExists(const char* path)
{
   const DWORD dwAttrib = GetFileAttributes(path);

  return (dwAttrib != INVALID_FILE_ATTRIBUTES && !(dwAttrib & FILE_ATTRIBUTE_DIRECTORY));
}
```

JeffMill: (with the generic name of "FileExists" or the like rather than "EnsurePayloadExists" as this method has nothing to do with "payload")

src/extensions/step_handlers/wim_handler/src/wim_step_handler.cpp
```diff
@@ -39,6 +41,76 @@ static std::string make_preferred(const char* path, const char* file = nullptr)
...
+
+/**
+ * @brief Get the volume label for the drive
+ *
+ * @param[in] driveRoot The drive root path name, e.g. "C:\\" or "D:\\"
+ * @param[out] outLabel The pointer to an existing std::string object to receive the resulting volume label.
+ * @return HRESULT The value of GetLastError() as processed by HRESULT_FROM_WIN32 macro on error, or S_OK on success.
+ * @details For failed returned HRESULT, the std::string object pointed to by outLabel will not be affected.
+ */
+static HRESULT GetDriveLabel(const std::string& driveRoot, std::string* outLabel)
```
JeffMill, line 70: seeing how line 86 can throw, consider throwing on error and returning a string, e.g.

``` C++
std::string GetDriveLabel(root) {
if (!GetVolumeInformation...)
{
  throw HRESULT_FROM_WIN32(GetLAstError();
}

return volumename;

File: src/extensions/step_handlers/wim_handler/src/wim_step_handler.cpp [hunk 5]
```diff
@@ -39,6 +41,76 @@ static std::string make_preferred(const char* path, const char* file = nullptr)
...
+}
+
+/**
+ * @brief Verifies that the expected volume label exists on the given drive.
+ *
+ * @param driveRoot The drive root, such as "C:\\" or "D:\\".
+ * @param expectedVolumeLabel The expected volume label.
+ * @return HRESULT S_OK when the give volume label matches that of the drive; an error HRESULT, otherwise.
+ */
+static HRESULT EnsureExpectedDriveLabel(const std::string& driveRoot, const std::string& expectedVolumeLabel)
```
JeffMill, line 97: here too, would a bool return be better and throw on error?

JeffMill (approved):
approved with comments.

JeffMill (approved):

PR #481, opened 2023-05-23
Fix WimHandler1 auto-reboot 

jw-msft:


src/extensions/step_handlers/wim_handler/src/wim_handler_1.cpp
```diff
@@ -289,6 +289,7 @@ ADUC_Result WimHandler1::Apply(const ADUC_WorkflowData* workflowData)
     else if (erc == WimStepHandler::RC::Success_Reboot_Required)
     {
         result.Set(ADUC_Result_Apply_RequiredReboot, 0);
```
JeffMill, line 291: there's also:

* ADUC_Result_Install_RequiredImmediateAgentRestart
* ADUC_Result_Install_RequiredImmediateReboot

do those cases need to be handled as well, or is RequiredReboot sufficient for now?  (I believe I asked Nox about this, but can't recall the answer)

JeffMill (approved):

PR #489, opened 2023-06-10
report ERCs str, rootkey workflow/dnld fixes, add rootkey telemetry

jw-msft:
- Add include_test_root_keys option in build.sh to make embedding test root keys explicit instead of implicitly part of Debug build (as opposed to non-Debug that will not contain test keys)
- Report "extendedResultCodes" JSON string (plural) instead of singular JSON number
- Add ADUC_ERC_ROOTKEY_PKG_UNCHANGED (0xa0000fff) for when downloaded rootkey pkg not changed from local store
- Address TODOs by adding more ERCs in rootkey workflow including common infra ERC
- Handle RootKey Pkg download case when pkg json file already exists in sandbox
- Fix rootkey workflow issues
- Add unit tests


src/agent/adu_core_interface/src/adu_core_interface.c
```diff
@@ -613,6 +608,21 @@ static _Bool set_workflow_properties(
     return succeeded;
 }
 
+static STRING_HANDLE construct_extended_result_codes_str(ADUC_WorkflowHandle handle, ADUC_Result rootResult)
+{
+    STRING_HANDLE root_result_erc_str =
+        ADUC_ReportingUtils_CreateReportingErcHexStr(rootResult.ExtendedResultCode, true /* is_first */);
+    STRING_HANDLE extra_ercs_str = workflow_get_extra_ercs(handle);
```
Nox-MSFT, line 615: Check whether root_result_erc_str == null here, and bail early to avail unnecessary work?

Nox-MSFT (approved):
Approved with minor comment.

src/agent/adu_core_interface/src/adu_core_interface.c
jw-msft: That is already checked on line 616 in 3rd and 4th terms of "and" conditional expr:
 if (extra_ercs_str != NULL && STRING_length(extra_ercs_str) > 0 _**&& root_result_erc_str != NULL
        && STRING_length(root_result_erc_str) > 0**_)

PR #490, opened 2023-06-14
Support Ubuntu 22.04 arm64 amd64

shiyi-peng:


scripts/install-deps.sh
```diff
@@ -164,10 +186,12 @@ do_install_aduc_packages() {
     # The latest version of gcc available on Debian is gcc-6. We install that version if we are
     # building for Debian, otherwise we install gcc-8 for Ubuntu.
     OS=$(lsb_release --short --id)
-    if [[ $OS == "debian" && $VER == "9" ]]; then
+    if [[ $OS == "Debian" && $VER == "9" ]]; then
         $SUDO apt-get install --yes gcc-6 g++-6 || return
     elif [[ $OS == "Debian" && $VER == "11" ]]; then
         $SUDO apt-get install --yes gcc-10 g++-10 || return
+    elif [[ $OS == "Ubuntu" && $VER == "22.04" ]]; then
+        $SUDO apt-get install --yes gcc-10 g++-10 || return
```
jw-msft, line 194: Could we change to not install particular one so that we do not have to maintain logic of mappings between distro + distro version?  
Just install the one that APT chooses?
e.g. 
```sh
$SUDO apt-get install --yes gcc g++
```

scripts/install-deps.sh
```diff
@@ -182,6 +206,12 @@ do_install_aduc_packages() {
 
     # Note that clang-tidy requires clang to be installed so that it can find clang headers.
     $SUDO apt-get install --yes "${static_analysis_packages[@]}" || return
+
+    # Check if the directory exists
+    if [ ! -d "$openssl_dir_path" ]; then
```
jw-msft, line 211: I believe Ubuntu 22.04 uses OpenSSL 3 for the system. Can we instead use newer version of azure-iot-hub-sdk for C that uses OpenSSL 3 when OpenSSL 3 is detected instead of installing OpenSSL 1.1.1 SxS with OpenSSL 3?

scripts/install-deps.sh
```diff
@@ -456,6 +487,7 @@ do_install_do() {
     local do_cmake_options=(
         "-DDO_BUILD_TESTS:BOOL=OFF"
         "-DDO_INCLUDE_SDK=ON"
+        "-DOPENSSL_ROOT_DIR=$work_folder/deviceupdate-openssl"
```
jw-msft, line 490: Does DO only work with OpenSSL 1.x or can it work with OpenSSL 3.x?
Same question for ABSFUU below.

scripts/install-deps.sh
```diff
@@ -77,9 +79,10 @@ install_do=false
 do_ref=$default_do_ref
 
 # Dependencies packages
-aduc_packages=('git' 'make' 'build-essential' 'cmake' 'ninja-build' 'libcurl4-openssl-dev' 'libssl-dev' 'uuid-dev' 'python2.7' 'lsb-release' 'curl' 'wget' 'pkg-config')
+# removed  'libssl-dev'
+aduc_packages=('git' 'make' 'build-essential' 'cmake' 'ninja-build' 'libcurl4-openssl-dev' 'uuid-dev' 'python2.7' 'lsb-release' 'curl' 'wget' 'pkg-config')
 static_analysis_packages=('clang' 'clang-tidy' 'cppcheck')
-compiler_packages=("gcc-[68]")
+compiler_packages=("gcc-{6,8,12}")
```
jw-msft, line 85: maybe this can just be: `=('gcc' 'g++')`

scripts/install-deps.sh
```diff
@@ -156,6 +159,25 @@ do_install_githooks() {
...
+do_install_openssl() {
+    local openssl_dir=$work_folder/openssl-1.1.1f
+    echo "Installing OpenSSL ..."
+
+    $SUDO apt-get install -y build-essential checkinstall zlib1g-dev -y || return
+    wget https://www.openssl.org/source/openssl-1.1.1f.tar.gz || return
+    mkdir -p $openssl_dir || return
+    tar -xf openssl-1.1.1f.tar.gz -C $openssl_dir --strip-components=1 || return
+    pushd $openssl_dir > /dev/null || return
+    ./config --prefix="$openssl_dir_path" --openssldir="$openssl_dir_path" shared zlib || return
```
jw-msft, line 171: openssl_dir_path is based on /tmp workfolder, so the shared .so lib files will be in /tmp which is typically mounting a tmpfs filesystem (runs only in RAM).  After a reboot, these .so library will be deleted.  Therefore, I recommend adding `--openssl-dir-path` argument to install-deps.sh to allow script user to specify a more permanent location to save the openssl 1.x .so files.

jw-msft (changes requested):

scripts/install-deps.sh
shiyi-peng: Yeah I was told that across Microsoft projects including the iot c sdk, they decide not to move to OpenSSL 3.0

scripts/install-deps.sh
shiyi-peng: sounds good, let me do it when adding support for debian 11 this quarter

azurepipelines/build/native/adu-ubuntu-amd64-build.yml
```diff
@@ -60,6 +60,17 @@ pr:
             - licenses/*
 
 jobs:
+    - job: BuildAduAgent_ubuntu2204
+      displayName: "Build ADU Agent - Ubuntu 22.04 (amd64)"
+      timeoutInMinutes: 60
```
jw-msft, line 65: these are 60 in this file but the other files are 120. Should these be 120 too, or reason to keep at 60?

scripts/install-deps.sh
```diff
@@ -77,9 +79,10 @@ install_do=false
 do_ref=$default_do_ref
 
 # Dependencies packages
+# removed  'libssl-dev'
```
jw-msft, line 82: still seems to be in aduc_packages list.

scripts/install-deps.sh
```diff
@@ -216,6 +249,7 @@ do_install_azure_iot_sdk() {
         "-Dskip_samples:BOOL=ON"
         "-Dbuild_service_client:BOOL=OFF"
         "-Dbuild_provisioning_service_client:BOOL=OFF"
+        "-DOPENSSL_ROOT_DIR=$work_folder/deviceupdate-openssl"
```
jw-msft, line 252: Should this be changed to $openssl_dir_path?

scripts/install-deps.sh
```diff
@@ -456,6 +490,7 @@ do_install_do() {
     local do_cmake_options=(
         "-DDO_BUILD_TESTS:BOOL=OFF"
         "-DDO_INCLUDE_SDK=ON"
+        "-DOPENSSL_ROOT_DIR=$work_folder/deviceupdate-openssl"
```
jw-msft, line 493: change to $openssl_dir_path?

scripts/install-deps.sh
```diff
@@ -514,6 +549,8 @@ do_install_azure_blob_storage_file_upload_utility() {
         azure_blob_storage_file_upload_utility_cmake_options+=("-DCMAKE_BUILD_TYPE:STRING=Release")
     fi
 
+    azure_blob_storage_file_upload_utility_cmake_options+=("-DOPENSSL_ROOT_DIR=$work_folder/deviceupdate-openssl")
```
jw-msft, line 552: change to $openssl_dir_path?

scripts/build.sh
```diff
@@ -366,7 +373,31 @@ else
...
+    $SUDO apt-get install -y build-essential checkinstall zlib1g-dev -y || return
+    wget https://www.openssl.org/source/openssl-1.1.1f.tar.gz || return
+    mkdir -p $openssl_dir || return
+    tar -xf openssl-1.1.1f.tar.gz -C $openssl_dir --strip-components=1 || return
+    pushd $openssl_dir > /dev/null || return
+    ./config --prefix="$openssl_dir_path" --openssldir="$openssl_dir_path" shared zlib || return
+    make || return
+
+    $SUDO make install || return
+    export LD_LIBRARY_PATH="$work_folder/deviceupdate-openssl/lib:$LD_LIBRARY_PATH" || return
```
jw-msft, line 389: change to $openssl_dir_path and in echo?

scripts/install-deps.sh
```diff
@@ -156,6 +161,25 @@ do_install_githooks() {
     fi
 }
 
+do_install_openssl() {
+    local openssl_dir=$work_folder/openssl-1.1.1f
```
jw-msft, line 165: why 1.1.1f?  There were many CVE vulnerabilities fixed since `f` as can be seen here: https://www.openssl.org/news/vulnerabilities-1.1.1.html

I think it should be latest to get all the security fixes, which is `1.1.1u`
https://www.openssl.org/source/openssl-1.1.1u.tar.gz

scripts/install-deps.sh
```diff
@@ -156,6 +161,25 @@ do_install_githooks() {
...
+    echo "Installing OpenSSL ..."
+
+    $SUDO apt-get install -y build-essential checkinstall zlib1g-dev -y || return
+    wget https://www.openssl.org/source/openssl-1.1.1f.tar.gz || return
+    mkdir -p $openssl_dir || return
+    tar -xf openssl-1.1.1f.tar.gz -C $openssl_dir --strip-components=1 || return
```
jw-msft, line 171: move `openssl-1.1.1f` to variable at top of file (should be `openssl-1.1.1u` I think) so can easily update the openssl version when new vulnerabilities are fixed in `u` and need to move to a future `v`

jw-msft (approved):
🚢

PR #494, opened 2023-06-17
Enabling Customer Overrides for Default Paths

Nox-MSFT:
#  Enabling Customer Overrides for Default Paths
>Note | Internal tracking task #43366149 & #43150953

This pull request enables customers to customize default paths by specifying a custom path in the du-config.json file.

The following build time configurations (compiler directives) can be overridden by specifying desired values in the du-config.json file:

| Variable | du-config property name | Default value | Override value |
|---|---|---|---|
| ADUC_DATA_FOLDER | dataFolder | /var/lib/adu | {dataFolder} |
| ADUC_DOWNLOADS_FOLDER | downloadsFolder | /var/lib/adu/downloads | {downloadsFolder} <br/> or fallback to {dataFolder}/downloads <br/> or fallback to default value |
| ADUC_EXTENSIONS_FOLDER | extensionsFolder | /var/lib/adu/extensions | {extensionsFolder} <br/>or fallback to {dataFolder}/extensions <br/> or fallback to default value |
| ADUC_COMPONENT_ENUMERATOR_EXTENSION_DIR | N/A | /var/lib/adu/extensions/component_enumerator | {extensionsFolder}/component_enumerator |
| ADUC_UPDATE_CONTENT_HANDLER_EXTENSION_DIR | N/A | /var/lib/adu/extensions/update_content_handlers | {extensionsFolder}/update_content_handlers |
| ADUC_CONTENT_DOWNLOADER_EXTENSION_DIR | N/A | /var/lib/adu/extensions/content_downloader | {extensionsFolder}/content_downloader |
| ADUC_DOWNLOAD_HANDLER_EXTENSION_DIR | N/A | /var/lib/adu/extensions/download_handlers | {extensionsFolder}/download_handlers |
| ADUSHELL_FOLDER | aduShellFolder | /usr/lib/adu | {aduShellFolder} |
| ADUSHELL_FILE_PATH | N/A | /usr/lib/adu/adu-shell | {aduShellFolder}/adu-shell |

## Specifying a custom du-config.json path
You can specify a custom du-config.json folder by using the aducIotAgent command-line options "-F" or "--config-folder".

For example:
```sh
# Load du-config.json from /user/local/adu/du-config.json
# instead of the default: /etc/adu/du-config.json
aducIotAgent -F "/user/local/adu"
```

To ensure that adu-shell also loads configurations from the same location, the same du-config folder will be passed to adu-shell at runtime.

By default, adu-shell will be invoked with "-F /etc/adu". If aducIotAgent is invoked with "-F /user/local/adu", this path value will be passed along to adu-shell.

## Change adu-shell default path
This pull request also change the adu-shell path from /usr/lib/adu to /usr/bin


yuvalg-MSFT (approved):

src/utils/config_utils/src/config_utils.c
```diff
@@ -347,14 +523,30 @@ bool ADUC_ConfigInfo_Init(ADUC_ConfigInfo* config, const char* configFilePath)
...
+    free(config->extensionsFolder);
+    free(config->extensionsComponentEnumeratorFolder);
+    free(config->extensionsContentDownloaderFolder);
+    free(config->extensionsStepHandlerFolder);
+    free(config->extensionsDownloadHandlerFolder);
+    free(config->iotHubProtocol);
     free(config->manufacturer);
     free(config->model);
-    free(config->iotHubProtocol);
+
```
jw-msft, line 549: why removing free of iotHubProtocol?

src/utils/config_utils/src/config_utils.c
```diff
@@ -448,3 +640,68 @@ void ADUC_ConfigInfo_FreeAduShellTrustedUsers(VECTOR_HANDLE users)
...
+ * @brief Get the existing ADUC_ConfigInfo object, or create one if it doesn't exist.
+ *
+ * @return const ADUC_ConfigInfo* a pointer to ADUC_ConfigInfo object. NULL if failure.
+ * Caller must call ADUC_ConfigInfo_Release to free the object.
+ */
+const ADUC_ConfigInfo* ADUC_ConfigInfo_GetInstance()
+{
+    const ADUC_ConfigInfo* config = NULL;
+
+    s_config_lock();
```
jw-msft, line 654: Why lock and refCount is needed? Trying to save memory? Why can't it just read config into global obj once during startup on main thread? then do not need lock / refcount for multiple threads to read the data.  Is profiling show that it is large consumer of memory?

src/utils/config_utils/src/config_utils.c
```diff
@@ -330,9 +389,126 @@ bool ADUC_ConfigInfo_Init(ADUC_ConfigInfo* config, const char* configFilePath)
...
+
+    if (NULL
+        == (config->extensionsComponentEnumeratorFolder =
+                ADUC_StringFormat("%s/%s", config->extensionsFolder, ADUC_EXTENSIONS_SUBDIR_COMPONENT_ENUMERATOR)))
+    {
+        Log_Error("Failed to allocate memory for component enumerator extension folder");
+        goto done;
+    }
+
+    if (NULL
```
jw-msft, line 479: Seems like it needs clang-format?

src/utils/config_utils/src/config_utils.c
```diff
@@ -330,9 +389,126 @@ bool ADUC_ConfigInfo_Init(ADUC_ConfigInfo* config, const char* configFilePath)
         goto done;
     }
 
+    const char* aduShellFolder = ADUC_JSON_GetStringFieldPtr(root_value, CONFIG_ADU_SHELL_FOLDER);
+
+    if (aduShellFolder != NULL)
+    {
+        if (mallocAndStrcpy_s(&(config->aduShellFolder), aduShellFolder) != 0)
```
jw-msft, line 396: why doing all this malloc and copy? Callers do not need clone of config file but can just share global obj. At very least, make clone_config() function.

src/platform_layers/linux_platform_layer/src/linux_adu_core_exports.cpp
```diff
@@ -68,14 +69,26 @@ void ADUC_Unregister(ADUC_Token token)
...
     sync();
 
-    std::string output;
-    std::vector<std::string> args{ "--update-type", "common", "--update-action", "reboot" };
-    const int exitStatus = ADUC_LaunchChildProcess(ADUSHELL_FILE_PATH, args, output);
+    const ADUC_ConfigInfo* config = ADUC_ConfigInfo_GetInstance();
+    if (config == NULL)
+    {
+        Log_Error("Failed to get config info instance.");
+        exitStatus = 126; // Command invoked cannot execute.
```
jw-msft, line 83: magic number. where is 126 coming from? it is supposed to be ENOKEY (126)?  https://github.com/torvalds/linux/blob/master/include/uapi/asm-generic/errno.h#LL110C9-L110C9
Should it be ENOMEM instead?  #include <errno.h> and use symbol

src/platform_layers/linux_platform_layer/src/linux_adu_core_exports.cpp
```diff
@@ -68,14 +69,26 @@ void ADUC_Unregister(ADUC_Token token)
...
+    }
+    else
+    {
+        const std::vector<std::string> args = { "--config-folder", config->configFolder, "--update-type",
```
jw-msft, line 87: run clang-format?

src/extensions/step_handlers/swupdate_handler/src/swupdate_handler.cpp
```diff
@@ -467,27 +492,46 @@ ADUC_Result SWUpdateHandlerImpl::IsInstalled(const tagADUC_WorkflowData* workflo
...
+            adushconst::target_log_folder_opt, logFolder
+        };
+
+        const int exitCode = ADUC_LaunchChildProcess(config->aduShellFilePath, aduShellArgs, output);
+        if (exitCode != 0)
+        {
+            // If failed to cancel apply, apply should return SuccessRebootRequired.
+            Log_Error("Failed to cancel Apply, extendedResultCode = %d", exitCode);
+            result.ResultCode = ADUC_Result_Failure;
+            result.ExtendedResultCode = ADUC_ERC_SWUPDATE_HANDLER_CANCEL_FAILED_TO_CANCEL_APPLIED_UPDATE;
```
jw-msft, line 524: Should it use MAKE_ERC... macro to pack the exitCode into lowest 20 bits into the corr facility + component?

src/extensions/step_handlers/swupdate_handler/src/swupdate_handler.cpp
```diff
@@ -278,38 +290,50 @@ ADUC_Result SWUpdateHandlerImpl::Install(const tagADUC_WorkflowData* workflowDat
...
 
-    if (exitCode != 0)
-    {
-        Log_Error("Apply failed, extendedResultCode = %d", exitCode);
-        result = { ADUC_Result_Failure, exitCode };
-        goto done;
+        if (exitCode != 0)
+        {
+            Log_Error("Apply failed, extendedResultCode = %d", exitCode);
+            result = { ADUC_Result_Failure, exitCode };
```
jw-msft, line 334: Should use MAKE_ADUC_EXTENDEDRESULTCODE_FOR_FACILITY_ADUC_FACILITY_UNKNOWN(ERRNO, exitCode)

src/extensions/step_handlers/swupdate_handler/src/swupdate_handler.cpp
```diff
@@ -278,38 +290,50 @@ ADUC_Result SWUpdateHandlerImpl::Install(const tagADUC_WorkflowData* workflowDat
...
 
     if (workflow_is_cancel_requested(workflowData->WorkflowHandle))
     {
         return this->Cancel(workflowData);
     }
 
+    config = ADUC_ConfigInfo_GetInstance();
+    if (config == nullptr)
+    {
+        Log_Error("Failed to get config info instance.");
```
jw-msft, line 304: Probably do not need log as this should never really be NULL in practice

src/extensions/step_handlers/swupdate_handler/src/swupdate_handler.cpp
```diff
@@ -226,6 +228,14 @@ ADUC_Result SWUpdateHandlerImpl::Install(const tagADUC_WorkflowData* workflowDat
         goto done;
     }
 
+    config = ADUC_ConfigInfo_GetInstance();
+    if (config == nullptr)
+    {
+        Log_Error("Failed to get config info instance.");
```
jw-msft, line 234: omit log, should never get here

src/extensions/step_handlers/script_handler/src/script_handler.cpp
```diff
@@ -535,6 +536,14 @@ ScriptHandler_PerformAction(const std::string& action, const tagADUC_WorkflowDat
         return results;
     }
 
+    const ADUC_ConfigInfo* config = ADUC_ConfigInfo_GetInstance();
+    if (config == nullptr)
+    {
+        Log_Error("Failed to get config info instance");
```
jw-msft, line 542: omit log to avoid adding string to the binary for something that will almost never happen

src/extensions/step_handlers/apt_handler/src/apt_handler.cpp
```diff
@@ -316,19 +330,22 @@ ADUC_Result AptHandlerImpl::Install(const ADUC_WorkflowData* workflowData)
         goto done;
     }
 
+    config = ADUC_ConfigInfo_GetInstance();
+    if (config == NULL)
+    {
+        Log_Error("Failed to get config instance.");
```
jw-msft, line 336: string keeps logging string keeps repeating. Log in ADUC_ConfigInfo_GetInstance() or omit, since it should not happen

jw-msft: line 186 and anywhere else calling ADUC_ConfigInfo_GetInstance

src/extensions/extension_manager/src/extension_manager_helper.cpp
```diff
@@ -120,19 +120,28 @@ ADUC_Result ProcessDownloadHandlerExtensibility(
...
  * @param downloadOptions The download options.
  * @return unsigned int The download timeout that should be used.
  */
 unsigned int GetDownloadTimeoutInMinutes(const ExtensionManager_Download_Options* downloadOptions) noexcept
 {
-    ADUC_ConfigInfo config{};
-    if (!ADUC_ConfigInfo_Init(&config, ADUC_CONF_FILE_PATH) || (config.downloadTimeoutInMinutes == 0))
+    unsigned int ret = CONTENT_DOWNLOADER_MAX_TIMEOUT_IN_MINUTES_DEFAULT;
+    const ADUC_ConfigInfo* config = ADUC_ConfigInfo_GetInstance();
+    if (config == nullptr)
```
jw-msft, line 131: probably better to crash if this is NULL, which it should never really be. systemd will restart anyways.

src/agent/src/health_management.c
```diff
@@ -469,42 +476,52 @@ static bool CheckShellBinary()
...
         // Note: other has no permission bits set.
         const mode_t expected_permissions = S_ISUID | S_IRUSR | S_IXUSR | S_IRGRP | S_IXGRP;
 
-        if (!PermissionUtils_VerifyFilemodeExact(path, expected_permissions))
+        if (!PermissionUtils_VerifyFilemodeExact(config->aduShellFilePath, expected_permissions))
         {
-            Log_Error("Lookup failed or '%s' has incorrect permissions (expected: 0%o)", path, expected_permissions);
+            Log_Error(
+                "Lookup failed or '%s' has incorrect permissions (expected: 0%o)",
+                config->aduShellFilePath,
```
jw-msft, line 509: why customer needs to override aduShell file path? Might be better to invest in installing all ADU files under different prefix e.g. /opt/my/path/{var/lib/adu/, etc/adu/, usr/bin/}

src/adu-shell/src/main.cpp
```diff
@@ -265,16 +277,16 @@ bool ADUShell_PermissionCheck()
     bool isTrusted = false;
 
     // If config file is provided, check if user is in trusted user list.
-    ADUC_ConfigInfo config = {};
-    if (ADUC_ConfigInfo_Init(&config, ADUC_CONF_FILE_PATH))
+    const ADUC_ConfigInfo* config = ADUC_ConfigInfo_GetInstance();
+    if (config != NULL)
```
jw-msft, line 281: if config == NULL, then caller needs to return ENOMEM rather than EPERM. Maybe can just remove NULL check. When would it happen?

packages/debian/postinst
```diff
@@ -20,7 +20,7 @@ adu_diagnostics_conf_file=du-diagnostics-config.json
 adu_log_dir=/var/log/adu
 adu_data_dir=/var/lib/adu
 adu_downloads_dir="$adu_data_dir/downloads"
-adu_shell_dir=/usr/lib/adu
+adu_shell_dir=/usr/bin
```
jw-msft, line 23: what if customer override shell path in du-config.json? Now, it is out of sync.

jw-msft (changes requested):

src/utils/config_utils/src/config_utils.c
Nox-MSFT: The main purpose of this change is to ensure that the configuration file is parsed only once, as needed, per process. Furthermore, as you've mentioned, if the main thread creates the file once and frees it upon exit, the other threads will not need to parse the configuration file again.

src/utils/config_utils/src/config_utils.c
Nox-MSFT: That's what caused by running clang-format....

src/utils/config_utils/src/config_utils.c
Nox-MSFT: It's moved to line 546

src/utils/config_utils/src/config_utils.c
Nox-MSFT: The new code follows the existing code pattern, where we copy strings from a JSON object (parson) to an ADUC_ConfigInfo object. We're not supposed to keep the root_value object around. It seems like there's a bug in the previous code where the root_value wasn't freed properly.
I'll free the root_value. Thanks for pointing that out.

src/extensions/step_handlers/apt_handler/src/apt_handler.cpp
Nox-MSFT: It's better to log this at the caller so that we can see the caller function name in the log file, instead of seeing  [AUC_ConfigInfoGetIntance]

src/extensions/extension_manager/src/extension_manager_helper.cpp
Nox-MSFT: I this case, if we're failing to get the download timeout value, we should fallback to the default value instead of crashing the agent. So that we don't see the agent keep crashing over and over.

packages/debian/postinst
Nox-MSFT: The assumption is that customers who modify the adu-shell portion are creating a customized agent and likely not installing it through the APT package.

src/extensions/extension_manager/src/extension_manager_helper.cpp
jw-msft: Yes, this makes sense in this case.

src/utils/config_utils/src/config_utils.c
jw-msft: Somehow I missed that!

src/utils/config_utils/src/config_utils.c
jw-msft: I see. I still think could call init and uninit in main func of each process, then a little simpler and don't need refcount / lock. refcount makes less sense here since refcount never expected to go to 0.

Nox-MSFT: As discussed off-line, this is a bit more practical when the agent forks many child processes that sometime need to load configs. Refcount make it simple to manages the config info lifecycle

src/utils/config_utils/src/config_utils.c
Nox-MSFT: I updated the code to avoid unnecessary memory allocation, and stash some value in the parson json object where they make sense. Let me know if you have any more concerns.

src/platform_layers/linux_platform_layer/src/linux_adu_core_exports.cpp
Nox-MSFT: Fix

src/platform_layers/linux_platform_layer/src/linux_adu_core_exports.cpp
Nox-MSFT: Ran clang-format.

src/utils/config_utils/src/config_utils.c
```diff
@@ -330,9 +389,126 @@ bool ADUC_ConfigInfo_Init(ADUC_ConfigInfo* config, const char* configFilePath)
...
+    if (downloadsFolder != NULL)
+    {
+        if (mallocAndStrcpy_s(&(config->downloadsFolder), downloadsFolder) != 0)
+        {
+            goto done;
+        }
+    }
+    else
+    {
+        config->downloadsFolder = ADUC_StringFormat("%s/%s", config->dataFolder, "downloads");
```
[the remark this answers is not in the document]
Nox-MSFT: Fix

src/utils/config_utils/src/config_utils.c
```diff
@@ -330,9 +389,126 @@ bool ADUC_ConfigInfo_Init(ADUC_ConfigInfo* config, const char* configFilePath)
...
+    if (extensionsFolder != NULL)
+    {
+        if (mallocAndStrcpy_s(&(config->extensionsFolder), extensionsFolder) != 0)
+        {
+            goto done;
+        }
+    }
+    else
+    {
+        config->extensionsFolder = ADUC_StringFormat("%s/%s", config->dataFolder, "extensions");
```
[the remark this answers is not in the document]
Nox-MSFT: Fix

src/utils/config_utils/src/config_utils.c
```diff
@@ -330,9 +389,126 @@ bool ADUC_ConfigInfo_Init(ADUC_ConfigInfo* config, const char* configFilePath)
...
+        if (config->extensionsFolder == NULL)
+        {
+            Log_Error("Failed to allocate memory for extension folder");
+            goto done;
+        }
+    }
+
+    if (NULL
```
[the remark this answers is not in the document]
Nox-MSFT: Fix

src/utils/config_utils/src/config_utils.c
```diff
@@ -298,13 +357,13 @@ bool ADUC_ConfigInfo_Init(ADUC_ConfigInfo* config, const char* configFilePath)
...
 
-    config->aduShellTrustedUsers = json_object_get_array(root_object, "aduShellTrustedUsers");
+    config->aduShellTrustedUsers = json_object_get_array(root_object, CONFIG_ADU_SHELL_TRUSTED_USERS);
     if (config->aduShellTrustedUsers == NULL)
     {
         goto done;
     }
 
-    const char* compatPropertyNames = ADUC_JSON_GetStringFieldPtr(root_value, "compatPropertyNames");
+    const char* compatPropertyNames = ADUC_JSON_GetStringFieldPtr(root_value, CONFIG_COMPAT_PROPERTY_NAMES);
```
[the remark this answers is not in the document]
Nox-MSFT: Fix

src/adu-shell/src/main.cpp
Nox-MSFT: The config can be null if du-config.json file is missing or not accessible, or any required field is missing.

src/extensions/step_handlers/swupdate_handler/src/swupdate_handler.cpp
Nox-MSFT: Fix. Returning ADUC_ERC_SWUPDATE_HANDLER_CHILD_FAILURE_PROCESS_EXITCODE

src/extensions/step_handlers/swupdate_handler/src/swupdate_handler.cpp
Nox-MSFT: Removed

src/extensions/step_handlers/swupdate_handler/src/swupdate_handler.cpp
Nox-MSFT: Removed

src/extensions/step_handlers/script_handler/src/script_handler.cpp
Nox-MSFT: Removed

jw-msft (approved):

PR #495, opened 2023-06-17
Add enforcement of disabledSigningKeys rootkey pkg

jw-msft:
Stacked PR on top of user/jw-msft/rootkey-ercs that has PR out here: https://github.com/Azure/iot-hub-device-update/pull/489

# Summary
- Add support for enforcing disabledSigningKeys in the root key package
- Memory leak fixes
- Break out JWSResult slightly to bubble up more specific issues in jws utils layer.


nihemstr (approved):
Approved

nihemstr (approved):

src/utils/c_utils/src/string_c_utils.c
```diff
@@ -443,3 +443,35 @@ _Bool MallocAndSubstr(char** target, char* source, size_t len)
...
+
+/**
+ * @brief Produces a new string where each character is replaced with output of @p mapFn call.
+ *
+ * @param src The source string (will not be mutated).
+ * @param mapFn The mapping function.
+ * @return char* The new string with characters mapped. Caller owns and must call free(). NULL is returned on failures.
+ * @details returns NULL if string is null or empty. e.g. char* mapped = ADUC_StringUtils_Map(str, tolower);
+ */
+char* ADUC_StringUtils_Map(const char* src, int (*mapFn)(int))
```
JeffMill, line 455: what does mapFn take an int and not a char?

src/utils/c_utils/src/string_c_utils.c
```diff
@@ -443,3 +443,35 @@ _Bool MallocAndSubstr(char** target, char* source, size_t len)
     *target = t;
     return true;
 }
+
+/**
+ * @brief Produces a new string where each character is replaced with output of @p mapFn call.
```
JeffMill, line 448: it seems like you're using this to send to tolower(), but there's already strlwr (I think that's it). why reinvent the wheel?

src/utils/crypto_utils/src/crypto_lib.c
```diff
@@ -441,3 +458,139 @@ void CryptoUtils_FreeCryptoKeyHandle(CryptoKeyHandle key)
...
+/**
+ * @brief Computes a SHA256 hash from bytes.
+ *
+ * @param buf The handle to the input bytes to be hashed.
+ * @return CONSTBUFFER_HANDLE The digest buffer handle. Caller will own it and need to call CONSTBUFFER_DecRef() to free.
+ */
+CONSTBUFFER_HANDLE CryptoUtils_CreateSha256Hash(const CONSTBUFFER_HANDLE buf)
+{
+    const size_t Sha256DigestNumBytes = 32;
+    unsigned char hashBuf[Sha256DigestNumBytes];
```
JeffMill, line 471: is this assumed not be to null terminated? does 32 accountfor null termination?

src/utils/crypto_utils/src/crypto_lib.c
```diff
@@ -441,3 +458,139 @@ void CryptoUtils_FreeCryptoKeyHandle(CryptoKeyHandle key)
...
+    memset(hashBuf, 0, sizeof(hashBuf));
+    const CONSTBUFFER* constbuf = CONSTBUFFER_GetContent(buf);
```
JeffMill, line 473: use more descriptive name than "constbuf"

src/utils/crypto_utils/src/crypto_lib.c
```diff
@@ -441,3 +458,139 @@ void CryptoUtils_FreeCryptoKeyHandle(CryptoKeyHandle key)
...
+        return NULL;
+    }
+
+    // Initialize OpenSSL digest context
+    if ((mdctx = EVP_MD_CTX_new()) == NULL)
+    {
+        goto done;
+    }
+
+    msg_digest = GetEvpMdFromShaAlg("sha256");
```
JeffMill, line 492: can return nullptr -- assumption is that next line will fail?

src/utils/crypto_utils/src/crypto_lib.c
```diff
@@ -441,3 +458,139 @@ void CryptoUtils_FreeCryptoKeyHandle(CryptoKeyHandle key)
...
+    }
+
+    // Hash the input data
+    if (1 != EVP_DigestUpdate(mdctx, constbuf->buffer, constbuf->size))
+    {
+        goto done;
+    }
+
+    // Finalize the digest and get the output
+    if (1 != EVP_DigestFinal_ex(mdctx, &hashBuf[0], NULL))
```
JeffMill, line 507: &hashBuf[0] is just hashBuf, no?

src/utils/crypto_utils/src/crypto_lib.c
```diff
@@ -441,3 +458,139 @@ void CryptoUtils_FreeCryptoKeyHandle(CryptoKeyHandle key)
...
+    {
+        goto done;
+    }
+
+    output_digest = (unsigned char*)malloc(Sha256DigestNumBytes);
+    if (output_digest != NULL)
+    {
+        memcpy(output_digest, &hashBuf[0], Sha256DigestNumBytes);
```
JeffMill, line 515: here too -- hashBuf

src/utils/crypto_utils/src/crypto_lib.c
```diff
@@ -441,3 +458,139 @@ void CryptoUtils_FreeCryptoKeyHandle(CryptoKeyHandle key)
...
+
+        out_handle = CONSTBUFFER_Create(output_digest, Sha256DigestNumBytes);
```
JeffMill, line 517: isn't there an API to alloc and assign a const buffer handle so that you dont need to separately do malloc/memcpy, or am i thinking of somethign else?

src/utils/crypto_utils/src/crypto_lib.c
```diff
@@ -441,3 +458,139 @@ void CryptoUtils_FreeCryptoKeyHandle(CryptoKeyHandle key)
...
+        goto done;
+    }
+
+    der_encoded_bytes = NULL;
+    der_length = i2d_RSAPublicKey(rsa, &der_encoded_bytes); // DER PKCS#1
+    if (der_length == 0)
+    {
+        goto done;
+    }
+    // der_length = i2d_RSA_PUBKEY(rsa, &der_encoded_bytes); // cert format?
```
JeffMill, line 581: Is this a TODO?

src/utils/jws_utils/src/jws_utils.c
```diff
@@ -396,6 +517,71 @@ JWSResult VerifyJWSWithSJWK(const char* jws)
...
+
+    sha256HashPubKey = CryptoUtils_CreateSha256Hash(pubkey);
+    if(pubkey == NULL)
+    {
+        result = JWSResult_HashPubKeyFailed;
+        goto done;
+    }
+
+    // See if the hash of public key is on the Disallowed.
+    for(int i=0; i<VECTOR_size(disabledHashOfPubKeysList); ++i)
```
JeffMill, line 556: nit: run this PR through clang-format.

JeffMill (approved):
approved with comments

src/utils/c_utils/src/string_c_utils.c
jw-msft: b/c all tolower() and other such functions in ctype.h take and return int. I think it's because getc() and getchar() can return -1 (EOF) int so it interops with that type-wise.

src/utils/c_utils/src/string_c_utils.c
jw-msft: I could be wrong, but strlwr (or _strlwr_s) is non-standard, only available in msft c compiler. It might be good to add to ADUCPAL to make it available to non-win platforms.

I don't see strlwr in [FreeBSD string.h](https://cgit.freebsd.org/src/tree/include/string.h?id=c58ff411bc5b19d8fa2ce65fd9d8fabd1b13392f) or in [linux string.h](https://github.com/torvalds/linux/blob/master/include/linux/string.h), but I do see them in msvc crt: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/strlwr-wcslwr?view=msvc-170

src/utils/crypto_utils/src/crypto_lib.c
jw-msft: Correct. It is not null-terminated string but buffer of exactly 32 _unsigned_ char bytes, so exactly 32 bytes.

src/utils/crypto_utils/src/crypto_lib.c
jw-msft: yes same. expressing in array context syntactically rather than as pointer.

src/utils/crypto_utils/src/crypto_lib.c
jw-msft: will remove in follow-up pr

src/utils/jws_utils/src/jws_utils.c
jw-msft: fixed with https://github.com/Azure/iot-hub-device-update/pull/489/commits/69c022e7af65368415a106ade88fcc87ea0a7015 on upstream topic branch

src/utils/crypto_utils/src/crypto_lib.c
jw-msft: Fixed in upstream topic branch: https://github.com/Azure/iot-hub-device-update/pull/489/commits/ac88823edb927123e4462b31ddd21dbf651e4e91

src/utils/crypto_utils/src/crypto_lib.c
jw-msft: It already does the copy into extra bytes allocated in the const buffer. I was being silly as it's unnecessary malloc+memcpy. Fixed in upstream with: https://github.com/Azure/iot-hub-device-update/pull/489/commits/d66ff0eb8c767556d9b4cb95ed00fefe97d017e6

src/utils/crypto_utils/src/crypto_lib.c
jw-msft: fixed in upstream here: https://github.com/Azure/iot-hub-device-update/pull/489/commits/a056dd8c10f49a46aa5cea9cb3a2a4a08a0c7d01

PR #499, opened 2023-06-27
Small fix on snap yaml for publishing to Ubuntu Core Store

shiyi-peng:
add DO as stage-snaps, so that it is one of the snap dependency
add allow-auto-connection: true for DO and AIS, so that it won't need the manual connection. and add allow-auto-connection  for account-control because he account-control interface in Snap provides an application with the ability to create, delete, and manage user accounts on a system.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -100,6 +100,8 @@ parts:
   du-agent:
     plugin: nil
     source: .
+    stage-snaps:
+      - deliveryoptimization-agent
     override-build: |
       ./scripts/install-deps.sh --work-folder /tmp --install-cmake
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -204,12 +206,11 @@ plugs:
     interface: content
     content: aziot-identity-service
     target: $SNAP_DATA
+    allow-auto-connection: true
 
   account-control:
     interface: account-control
-
-  hardware-observe:
```
Nox-MSFT, line 211: After removing this plug, please ensure that all DeviceInfo_* functions are still functioning properly.

Nox-MSFT (approved):

PR #505, opened 2023-07-12
update e2e test yaml to unblock pipelines+fix some typos

shiyi-peng:


nihemstr (approved):
Approved with temporary  move over to docker builds tracked by bug here: https://microsoft.visualstudio.com/OS/_workitems/edit/45521400

PR #514, opened 2023-08-03
Moved project to target name and added proper handling of targets

nihemstr:
- Changed the code base to properly build targets when necessary and projects in other cases. This should help with project vs target scoping (mainly defining what is being built into what ala the targets without confusing the scope of project vs library)


src/extensions/content_downloaders/curl_downloader/CMakeLists.txt
```diff
@@ -1,23 +1,22 @@
-project (curl_content_downloader)
-
+set (target_name curl_content_downloader)
```
Nox-MSFT, line 1: For all extension, I think we should keep them 'project' instead. Since we may (eventually) split the package. And we want to manage the .SO file version individually.

src/extensions/content_downloaders/deliveryoptimization_downloader/CMakeLists.txt
```diff
@@ -1,25 +1,23 @@
-project (deliveryoptimization_content_downloader)
-
+set (target_name deliveryoptimization_content_downloader)
```
Nox-MSFT, line 1: Keep it a Project?

src/utils/hash_utils/CMakeLists.txt
```diff
@@ -1,24 +1,23 @@
 cmake_minimum_required (VERSION 3.5)
 
-project (hash_utils)
+set (target_name hash_utils)
+add_library (${target_name} STATIC src/hash_utils.c)
+add_library (aduc::${target_name} ALIAS ${target_name})
 
-add_library (${PROJECT_NAME} STATIC src/hash_utils.c)
-add_library (aduc::${PROJECT_NAME} ALIAS ${PROJECT_NAME})
+target_include_directories (${target_name} PUBLIC inc)
```
Nox-MSFT, line 7: Remove

src/extensions/content_downloaders/curl_downloader/CMakeLists.txt
nihemstr: We can still split it into a *.so. That's a target property not a project property. A project is the sum of all of the buildable targets within it. 

If we want to break it out into a shared library we'd be changing this to either a separate folder/repo from this project if we were to build it. 

The only time you traditionally have nested projects is when the higher level projects have no need of the targets built under neath it. 

That's the difference between a project vs target. 

If we wanted to change this to a shared library we'd need to change the add_library type from STATIC to DYNAMIC and CMake will handle the rest. It'll even handle the linking between the libraries as we go.

Nox-MSFT (approved):

src/adu-shell/CMakeLists.txt
```diff
@@ -1,7 +1,5 @@
 cmake_minimum_required (VERSION 3.5)
 
-project (azure_iot_aduc_shell)
-
 set (target_name adu-shell)
```
jw-msft, line 3: Nit: (can be addressed in separate PR) The other targets use underscore instead of hyphen

jw-msft (approved):

PR #516, opened 2023-08-10
Fix installs-deps -a catch2 issue on ubuntu 20.04

jw-msft:
Building catch2 from sources fails on `#pragma GCC diagnostic ignored "-Wcomma-subscript"` due to using gcc-9 and g++-9 packages on Ubuntu 20.04

This change installs gcc-10 and g++-10 packages on Ubuntu 20.04 to resolve this and also exports CC and CXX env variables so that catch2 CMake build will use gcc-10 and g++-10 instead of gcc-9 and g++-9.

### DIFF
--- a/scripts/install-deps.sh
+++ b/scripts/install-deps.sh
@@ -78,6 +78,14 @@ default_do_ref=develop
 install_do=false
 do_ref=$default_do_ref
 
+# catch2 build
+#
+# used for dependencies like catch2 that will find system default
+# (e.g. g++-9) despite g++-10 installed, so use CC and CXX env
+# vars that CMake will honor.
+catch2_cc=""
+catch2_cxx=""
+
 # Dependencies packages
 aduc_packages=('git' 'make' 'build-essential' 'cmake' 'ninja-build' 'libcurl4-openssl-dev' 'libssl-dev' 'uuid-dev' 'python2.7' 'lsb-release' 'curl' 'wget' 'pkg-config')
 static_analysis_packages=('clang' 'clang-tidy' 'cppcheck')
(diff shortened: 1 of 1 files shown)


shiyi-peng (dismissed):
awesome!

scripts/install-deps.sh
```diff
@@ -190,10 +190,17 @@ do_install_aduc_packages() {
...
-    elif [[ $OS == "Debian" && $VER == "11" ]]; then
-        $SUDO apt-get install --yes gcc-10 g++-10 || return
-    elif [[ $OS == "Ubuntu" && $VER == "22.04" ]]; then
+    elif [[ ($OS == "Debian" && $VER == "11") || (\
+        $OS == "Ubuntu" && $VER == "20.04") || (\
+        $OS == "Ubuntu" && $VER == "22.04") ]] \
+            ; then
         $SUDO apt-get install --yes gcc-10 g++-10 || return
+
+        # used for dependencies like catch2 that will find system default
```
JeffMill, line 199: I'm not thrilled about us modifying user environment for our needs.  Is there no way for us to build catch2 and pass along "CC=" and "CXX=" on the cmake commandline (or however we build catch2)?

scripts/install-deps.sh
```diff
@@ -293,6 +300,7 @@ do_install_catch2() {
     mkdir cmake || return
     pushd cmake > /dev/null || return
 
+    export CC=/usr/bin/gcc-10 CXX=/usr/bin/g++-10
```
JeffMill, line 303: as above

JeffMill (approved):
approved with comments

scripts/install-deps.sh
jw-msft: I'll forgo the export and inline on the cmake line, which will not modify env externally

jw-msft: See https://github.com/Azure/iot-hub-device-update/pull/516/commits/05cb632c4add1136d241c7deb0829a942e54efa5

scripts/install-deps.sh
jw-msft: Good suggestion. Yes, have pushed a commit that does this now.

shiyi-peng (approved):

Issue #504, opened 2023-07-11
Azure Device Update over IoT Hub for RHEL device.


sandipanaich:
Hi,

I am trying to build the Azure Device Update over IoT Hub on RHEL OS as well. I understand that it is not currently supported by Microsoft officially. I was curious if I can build it with RHEL, since there are quite some offerings that we have on RHEL.

So here is the step which I followed on an RHEL. 
Firstly I ran the following steps to ensure the RHEL support is not available.

`
wget [https://packages.microsoft.com/config/rhel/8/packages-microsoft-prod.rpm -O packages-microsoft-prod.rpm
sudo yum install packages-microsoft-prod.rpm
rm packages-microsoft-prod.rpm
sudo yum update
sudo yum install deviceupdate-agent
`
Clearly, the last step failed because the support for RHEL is not available yet.
sudo yum install deviceupdate-agent

So I cloned and ran the steps the mentioned in documents to build the code.

`
git clone https://github.com/Azure/iot-hub-device-update.git
./scripts/build.sh -c
`
I noticed that the last step failed. I also tried the following commands, but they failed too.
`
sudo ./scripts/build.sh --type rpm --install --cmake-path .
sudo ./scripts/build.sh --type rpm --install --cmake-path /home/sandipan/iot-hub-device-update
`
The Output in each case is as follows, the error can be seen in the last 2 lines.

`
Building ADU Agent
* Clean build: false
* Documentation: false
* Platform layer: linux
* Trace target deps: false
* Content handlers: microsoft/swupdate,microsoft/apt,microsoft/simulator
* Build type: rpm
* Log directory: /var/log/adu
* Logging library: zlog
* Output directory: /home/honadmin/sandipan/iot-hub-device-update/scripts/../out
* Build unit tests: false
* Build packages: false
* CMake: ./bin/cmake
./scripts/build.sh: line 321: ./bin/cmake: No such file or directory
* CMake version:
* shellcheck: /tmp/deviceupdate-shellcheck
./scripts/build.sh: line 323: /tmp/deviceupdate-shellcheck: No such file or directory
* shellcheck version:
* Static analysis: (none)

Error: No './bin/cmake' file.
Error: CMake failed to generate Ninja build with exit code: 1
`
Since I am running the build for the first time, I was curious why was the first error, ie. Error: No './bin/cmake' file appearing.

Based on these attempts, I would like to request you to kindly share the correct steps to build the codebase for RHEL.
[...]

PR #520, opened 2023-08-12
Build DU agent from sources on RHEL 9.2 / Rocky Linux

jw-msft:
- Updates necessary to build DU Agent on RHEL 9.2 or its drop-in replacement Rocky Linux (in the spirit of CentOS)
Addresses github issue: https://github.com/Azure/iot-hub-device-update/issues/504


jw-msft:
Removing this in favor of creating RPM package on debian-based system.

Issue #526, opened 2023-08-31
ScriptHandler crashs ADU


shaeussler:
Bug in line https://github.com/Azure/iot-hub-device-update/blob/b71516d3e85f09db48c06980448e3a07d90d05e4/src/utils/process_utils/src/process_utils.cpp#L110 when count is 1024.

## Expected Behavior
ADU should not crash with a segmentation fault

## Current Behavior
ADU crash with a segmentation fault

## Steps to Reproduce
Provide a detailed set of steps to reproduce the bug.
1. Run ADU and using the ScriptHandler

### Device Information
* Host OS: Buildroot Linux
* Architecture: arm64
* Provisioning Method: TPM
*DU Agent Version (run 'sudo -u adu /usr/bin/AducIotAgent -v'): 1.0.2

## Logs
-

## Additional Information
The following code snippet on line https://github.com/Azure/iot-hub-device-update/blob/b71516d3e85f09db48c06980448e3a07d90d05e4/src/utils/process_utils/src/process_utils.cpp#L110 solves the bug:
```
        if(count < sizeof(buffer))
        {
            buffer[count] = 0;
        }
```

jw-msft:
Thanks for reporting this, @shaeussler.

I have a PR #527 out to address this in main branch that also fixes same issue I found in [string_c_utils LoadBufferWithFileContents](https://github.com/Azure/iot-hub-device-update/blob/b71516d3e85f09db48c06980448e3a07d90d05e4/src/utils/c_utils/src/string_c_utils.c#L136)

I'm also porting the fix to develop branch.

jw-msft:
@shaeussler, Her is the merged commit in [main](https://github.com/Azure/iot-hub-device-update/tree/main) that should fix this issue: https://github.com/Azure/iot-hub-device-update/commit/f9b180dede4a6ed362cdf809862a9ef12872db4d

shaeussler:
Thanks

PR #527, opened 2023-08-31
Fix read.2 count arg off by one nul term write to buffer

jw-msft:
read(2) can read `count` (3rd arg) bytes, so writing of null term at array index of "num bytes read" in the buffer could overrun the buffer by 1.
- Fix process_utils `ADU_LaunchChildProcess` to read buf size minus one to address issue reported by:  https://github.com/Azure/iot-hub-device-update/issues/526
- Fix another instance in string_c_utils LoadBufferWithFileContents

### DIFF
--- a/src/utils/c_utils/src/string_c_utils.c
+++ b/src/utils/c_utils/src/string_c_utils.c
@@ -92,12 +92,12 @@ bool ReadDelimitedValueFromFile(const char* fileName, const char* key, char* val
  * @brief Function that sets @p strBuffers to the contents of the file at @p filePath if the contents are smaller in size than the buffer
  * @param filePath path to the file who's contents will be read
  * @param strBuffer buffer which will be loaded with the contents of @p filePath
- * @param strBuffSize the size of the buffer
+ * @param strBuffSize the size of the buffer to hold file contents plus a null terminator.
  * @returns false on failure, true on success
  */
 bool LoadBufferWithFileContents(const char* filePath, char* strBuffer, const size_t strBuffSize)
 {
-    if (filePath == NULL || strBuffer == NULL || strBuffSize == 0)
+    if (filePath == NULL || strBuffer == NULL || strBuffSize < 2)
     {
         return false;
     }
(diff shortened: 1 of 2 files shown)


Nox-MSFT (approved):

Jerry-GeGarcia (approved):

src/utils/c_utils/src/string_c_utils.c
```diff
@@ -92,12 +92,12 @@ bool ReadDelimitedValueFromFile(const char* fileName, const char* key, char* val
  * @brief Function that sets @p strBuffers to the contents of the file at @p filePath if the contents are smaller in size than the buffer
  * @param filePath path to the file who's contents will be read
  * @param strBuffer buffer which will be loaded with the contents of @p filePath
- * @param strBuffSize the size of the buffer
+ * @param strBuffSize the size of the buffer to hold file contents plus a null terminator.
  * @returns false on failure, true on success
  */
 bool LoadBufferWithFileContents(const char* filePath, char* strBuffer, const size_t strBuffSize)
```
JeffMill, line 98: looking at usage patterns, we seem to always do:

```c
        if (!LoadBufferWithFileContents(config.edgegatewayCertPath, certificateString, ARRAY_SIZE(certificateString)))
        {
            Log_Error("Failed to read the certificate from path: %s", config.edgegatewayCertPath);
            goto done;
        }

        if (mallocAndStrcpy_s(&info->certificateString, certificateString) != 0)
```

so there's no reason this method shouldn't do an allocation -- it's much safer to do (pseudo-code, add necessary error checking):

```c
// Caller must call free on returned pointer.
char* LoadBufferWithFileContents(const char* filePath)
{
  size = stat(filePath)
  buf = alloc(size);
  read(filePath, buf, size)
  return buf
}
```

JeffMill (changes requested):

src/utils/c_utils/src/string_c_utils.c
JeffMill: and in the case above, just do:

```c
info->certificateString = LoadBufferWithFileContents(config.edgegatewayCertPath);
if (info->certificateString == NULL) 
{
    Log_Error...
```

Nox-MSFT: :thinking::  Could also use `mmap` to avoid the memory allocation? **Assuming the file content is static.**

PR #529, opened 2023-08-31
E2E pipeline failure: apt lock

shiyi-peng:
1. Check for both apt and dpkg processes.
2. Add a sleep command between the commands for a small delay.

### DIFF
--- a/azurepipelines/e2e_test/templates/e2e_test_vm_cleanup.yaml
+++ b/azurepipelines/e2e_test/templates/e2e_test_vm_cleanup.yaml
@@ -6,10 +6,12 @@ parameters:
       type: string
 steps:
     - script: |
-          while pgrep apt > /dev/null; do sleep 1; done; sudo apt update && sudo apt install curl
+          while pgrep -f 'apt|dpkg' > /dev/null; do sleep 1; done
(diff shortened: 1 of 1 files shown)


azurepipelines/e2e_test/templates/e2e_test_vm_cleanup.yaml
```diff
@@ -6,10 +6,12 @@ parameters:
       type: string
 steps:
     - script: |
-          while pgrep apt > /dev/null; do sleep 1; done; sudo apt update && sudo apt install curl
+          while pgrep -f 'apt|dpkg' > /dev/null; do sleep 1; done
```
jw-msft, line 9: Why is `-f` needed? My thinking is it's not needed and may even be harmful because `-f` (or `--full`) matches the entire command line (including cmd-line arguments) instead of just the process name.

Nox-MSFT: Suggestion: instead of just calling `sleep()`, consider printing out the command line for the `'apt' or 'dpkg'` process. This will help us identify the specific command causing the issue.

PR #540, opened 2023-09-26
Updating the device enrollment state management module to saves data in the state-store and fix bugs

Nox-MSFT:


src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/adu_agent_info_management.c
```diff
@@ -508,10 +507,10 @@ void OnMessage_ainfo_resp(
                 }
             }
 
-            s_agentInfoMgrState.workflowState = ADUC_AGENT_INFO_WORKFLOW_STATE_SUBSCRIBING;
+            ADUC_StateStore_GetCommunicationChannelHandle(
+                &s_agentInfoMgrState.publishTopic, &s_agentInfoMgrState.publishTopic);
```
jw-msft, line 511: both args are same pointer addr?

src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/adu_mqtt_client_module.c
```diff
@@ -380,14 +346,36 @@ static int ADUC_MQTT_Client_Module_Initialize_DoWork(ADUC_MQTT_CLIENT_MODULE_STA
...
         {
-            Log_Info("MQTT broker hostname is empty. Retry in %d seconds", DEFAULT_OPERATION_INTERVAL_SECONDS);
-            moduleState->initializeState = ADU_MQTT_CLIENT_MODULE_INITIALIZE_STATE_PARTIAL;
-            // TODO (nox-msft) : use retry_utils
-            moduleState->nextOperationTime = ADUC_GetTimeSinceEpochInSeconds() + DEFAULT_OPERATION_INTERVAL_SECONDS;
+            const char* hostNameFromDPS = ADUC_StateStore_GetMQTTBrokerHostname();
+            if (!IsNullOrEmpty(hostNameFromDPS))
+            {
+                Log_Info("Applying MQTT broker endpoint from DPS.'%s'", hostNameFromDPS);
+                moduleState->mqttSettings.hostname = strdup(hostNameFromDPS);
```
jw-msft, line 353: It will goto done on line 367 if strdup here, but shouldn't it not goto done and continue on to get externalDeviceId on line 371?

src/extensions/agent_modules/shared/state_store/src/agent_state_store.c
```diff
@@ -476,6 +481,31 @@ ADUC_STATE_STORE_RESULT ADUC_StateStore_SetDeviceId(const char* deviceId)
     return ADUC_STATE_STORE_RESULT_OK;
 }
 
+/**
+ * @brief Get the DU service instance name from the state store.
+ */
+const char* ADUC_StateStore_GetDeviceUpdateServiceInstance(void)
+{
+    sem_wait(&state_semaphore);
```
jw-msft, line 489: Is locking necessary -- are there more than one thread accessing the store? If there are, what are the 2 or more threads doing so?

Why semaphore? could it just use mutex or reader-writer lock instead?  I'd typically use semaphore if waiting for some shared number to reach a certain value, but not for just protecting access to shared data. I see it's achieving similar thing by sem_post when done with a section but it seems simpler to me to just use mutex or reader-writer lock.

PR #545, opened 2023-10-02
Suppressing connection health check and fixing invalid handlers in agent main.

Nox-MSFT:


src/agent/src/main.c
```diff
@@ -507,30 +509,33 @@ int main(int argc, char** argv)
 
     // Initialize the agent modules.
     // TODO (nox-msft) - replace with Agent Module manager code, once changed all modules to shared library.
-    dpsClientModuleInterface = (ADUC_AGENT_MODULE_INTERFACE*)ADPS_MQTT_Client_Module_Create();
+    dpsClientModuleHandle = ADPS_MQTT_Client_Module_Create();
```
jw-msft, line 512: I'd argue that dpsClientModuleInterface should get assigned result of ADPS_MQTT_Client_Module_Create() that returns ADUC_AGENT_MODULE_INTERFACE* type.  Then, dspClientModuleHandle can cast it to ADUC_AGENT_MODULE_HANDLE that is a void* opaque type.

jw-msft: However, can pass the interface pointer directly into the methods such as `initializeModule` that take an `ADUC_AGENT_MODULE_HANDLE` since it will be coerced to `void*` implicitly.

jw-msft (approved):
Approved with suggestions.

PR #562, opened 2023-10-31
merge from develop

jw-msft:


src/CMakeLists.txt
```diff
@@ -1,5 +1,9 @@
 cmake_minimum_required (VERSION 3.5)
 
+include (agentRules)
+
+compileasc99 ()
```
nihemstr, line 5: Important Note: we don't need to ever state this again. As long as the head file is stating which standard to use all child targets will inherit it. 

The only places we need to specify compileasc99() is in new projects (not sure if that effects existing code within the POC)

src/inc/aduc/aduc_banned.h
```diff
@@ -0,0 +1,18 @@
...
+ * unsafe to result in a compilation error.
+ *
+ * @copyright Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT License.
+ */
+
+#ifndef ADUC_BANNED_H
+#define ADUC_BANNED_H
+
+#define DEFINE_BANNED_ALTERNATIVE(BANNED_FUNC, ALT_FUNC) BANNED_FUNC##_is_banned__Please_use_##ALT_FUNC##_instead
```
nihemstr, line 13: nit: Function name should be "BAN_HAMMER_TO_ALT()" because it would make me laugh.

src/utils/crypto_utils/CMakeLists.txt
```diff
@@ -9,6 +9,10 @@ target_include_directories (${target_name} PUBLIC inc)
 
 find_package (OpenSSL REQUIRED)
 
+message(OPENSSL_CRYPTO_LIBRARY="${OPENSSL_CRYPTO_LIBRARY}")
```
nihemstr, line 12: We need to delete these. I messed up and didn't remove them in the initial commit (it's useful for tracking across different plats but we should just do it once). 

Might be good to move to the project definition instead of here in a target definition (will only print once).

src/utils/crypto_utils/src/crypto_lib.c
```diff
@@ -13,6 +13,11 @@
 #include <ctype.h>
 #include <openssl/bn.h>
 #include <openssl/evp.h>
+
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
```
nihemstr, line 17: It's so ugly... So very, very ugly. 

No recommendation just wanted that in writing.

src/utils/crypto_utils/tests/CMakeLists.txt
```diff
@@ -10,10 +10,16 @@ disablertti ()
 set (sources main.cpp crypto_utils_ut.cpp)
 
 find_package (Catch2 REQUIRED)
+find_package (OpenSSL REQUIRED)
 
 add_executable (${PROJECT_NAME} ${sources})
 
-target_link_libraries (${PROJECT_NAME} PRIVATE aduc::crypto_utils aduc::string_utils Catch2::Catch2)
+message(OPENSSL_CRYPTO_LIBRARY="${OPENSSL_CRYPTO_LIBRARY}")
```
nihemstr, line 17: Need to remove. Not needed and already stated by crypto_utils.

src/utils/jws_utils/tests/CMakeLists.txt
```diff
@@ -12,6 +12,10 @@ set (sources main.cpp jws_utils_ut.cpp)
 find_package (Catch2 REQUIRED)
 find_package (OpenSSL REQUIRED)
 
+message(OPENSSL_CRYPTO_LIBRARY="${OPENSSL_CRYPTO_LIBRARY}")
+message(OPENSSL_VERSION="${OPENSSL_VERSION}")
```
nihemstr, line 16: Need to remove. Already printed out by crypto_utils.

tools/openssl_install_script/install-openssl.sh
```diff
@@ -0,0 +1,158 @@
+#!/bin/bash
+
```
nihemstr, line 2: Technically this is a helper script. Not sure if you want it in Gen2 when we can prescribe versioning (basically require OpenSSL 3.0 to run).

nihemstr (approved):
Approved with some minor nits about removing unnecessary output, adding in openssl script, and some other comments.

src/inc/aduc/aduc_banned.h
jw-msft: I like it but won't fix

tools/openssl_install_script/install-openssl.sh
jw-msft: In later PR, will remove

src/utils/crypto_utils/tests/CMakeLists.txt
jw-msft: will send a separate PR for these.

src/CMakeLists.txt
jw-msft: we also do not need to state `cmake_minimum_required (VERSION 3.5)` anywhere else either, I'm pretty sure.

src/utils/crypto_utils/CMakeLists.txt
jw-msft: done in new pr branch

PR #567, opened 2023-11-06
Adds the ability to create a service e2e compliant test agent

nihemstr:
- Adds a new build type for the agent that allows us to build the agent with test keys


src/utils/c_utils/src/string_c_utils.c
```diff
@@ -482,7 +482,7 @@ bool MallocAndSubstr(char** target, char* source, size_t len)
  * @return char* The new string with characters mapped. Caller owns and must call free(). NULL is returned on failures.
  * @details returns NULL if string is null or empty. e.g. char* mapped = ADUC_StringUtils_Map(str, tolower);
  */
-char* ADUC_StringUtils_Map(const char* src, int (*mapFn)(int))
+char* ADUC_StringUtils_Map(const char* src, char (*mapFn)(int))
```
jw-msft, line 485: Please revert this, or was it causing an issue? The functions in ctype.h that do side-effects with chars (tolower, toupper, etc) always accept and return `int` type since EOF is an int (usually -1).  The spec says the expected range of numbers is that of unsigned char type (0-255) or EOF.

Changing to `char` here is incorrect because mapFn will return EOF (-1) and on ARM platform, `char` type is always _unsigned_ so EOF will be converted to 255 but that wouldn't happen if mapFn is designated to return `int`.

jw-msft:
pipeline has 4 tests failing:

>The following tests FAILED:
	 62 - RSA Keys (Failed)
	 63 - Signature Verification (Failed)
	 78 - VerifySJWK (Failed)
	 79 - VerifyJWSWithKey (Failed)

src/utils/c_utils/src/string_c_utils.c
jw-msft: A good explanation of EOF and why it always needs to be int type: https://faq.cprogramming.com/cgi-bin/smartfaq.cgi?id=1043284351&answer=1048865140

jw-msft (approved):

PR #574, opened 2023-11-21
Removing erroneous removal of libxml2-dev from source dependencies

nihemstr:


scripts/install-deps.sh
```diff
@@ -11,6 +11,7 @@
 # Ensure that getopt starts from first option if ". <script.sh>" was used.
 OPTIND=1
 
+ret=""
```
Jerry-GeGarcia, line 14: ?

Jerry-GeGarcia (approved):
:shipit:

scripts/install-deps.sh
nihemstr: It needed to be declared so it was at a larger scope I think?

PR #575, opened 2023-11-21
Fixed issue where the work folder was not properly being set

nihemstr:
- get_work_folder_for_workflow() makes the assumption that ID would always be set which was causing the returned file path to end with '(null)' which when nested by the MCU messed with file pathing
- Added get_workflow_base_dir() so we have the ability to get the directory within which the work folder is going to be created no matter what
- Made the tools thing executable because it's a shell script
- Ran clang-format on result.h


src/agent/adu_core_interface/src/adu_core_interface.c
```diff
@@ -446,16 +446,17 @@ void OrchestratorUpdateCallback(
...
-        if (!ReportPreDeploymentProcessingState(propertyValue, ADUCITF_State_DeploymentInProgress, workflowData, inProgressResult))
+        if (!ReportPreDeploymentProcessingState(
+                propertyValue, ADUCITF_State_DeploymentInProgress, workflowData, inProgressResult))
         {
             Log_Warn("Reporting InProgress failed. Continuing processing deployment %s", workflowId);
         }
 
         // Ensure update to latest rootkey pkg, which is required for validating the update metadata.
-        workFolder = workflow_get_workfolder(workflowData->WorkflowHandle);
+        workFolder = workflow_get_workflow_base_dir(workflowData->WorkflowHandle);
```
JeffMill, line 456: what's the difference between "..._workfolder" and "..._workflow_base_dir"?  Seems like these can be (and were) easily confused?

src/inc/aduc/result.h
```diff
@@ -116,36 +116,47 @@ typedef enum tagADUC_FACILITY_UNKNOWN_Components
 
 typedef enum tagADUC_FACILITY_LOWERLAYER_Components
```
JeffMill, line 117: are there any changes in this file, or is this just the result of running clang-format on this file?

src/utils/rootkeypackage_utils/src/rootkeypackage_do_download.cpp
```diff
@@ -13,6 +13,8 @@
 
 EXTERN_C_BEGIN
 
+#define MAX_TIME_TO_DOWNLOAD_IN_SECS 1 /* hour*/ * 60 /* min*/ * 60 /* sec*/
```
JeffMill, line 16: re-read the email I sent on std::chrono :-)  If you have access to it, you shouldn't need to do any math.

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1804,57 +1804,86 @@ bool workflow_set_sandbox(ADUC_WorkflowHandle handle, const char* sandbox)
     return true;
 }
 
-// Workfolder =  [root.sandboxfolder]  "/"  ( [parent.workfolder | parent.id]  "/" )+  [handle.workfolder | handle.id]
-char* workflow_get_workfolder(const ADUC_WorkflowHandle handle)
+// Returns the download base directory
```
JeffMill, line 1807: download base directory, or workflow base directory?  this line and the method name conflict.

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1804,57 +1804,86 @@ bool workflow_set_sandbox(ADUC_WorkflowHandle handle, const char* sandbox)
...
-    if (wf != NULL)
-    {
-        Log_Debug("Property '%s' not NULL - returning cached workfolder '%s'", WORKFLOW_PROPERTY_FIELD_WORKFOLDER, wf);
-        free(id);
-        return wf;
-    }
+    char* pwf = NULL;
 
-    // Return ([parent's workfolder] or [default sandbox folder]) + "/" + [workflow id];
     ADUC_WorkflowHandle p = workflow_get_parent(handle);
```
JeffMill, line 1814: what does workflow_get_parent do?  parent of what?

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1804,57 +1804,86 @@ bool workflow_set_sandbox(ADUC_WorkflowHandle handle, const char* sandbox)
...
     if (p != NULL)
     {
         pwf = workflow_get_workfolder(p);
-        sprintf(dir, "%s/%s", pwf, id);
+        sprintf(dir, "%s/", pwf);
```
JeffMill, line 1818: you're asking for a stock overflow here and a zero-day. This isn't safe.

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1804,57 +1804,86 @@ bool workflow_set_sandbox(ADUC_WorkflowHandle handle, const char* sandbox)
...
     }
     else
     {
         const ADUC_ConfigInfo* config = ADUC_ConfigInfo_GetInstance();
         if (config != NULL)
         {
-            Log_Info("Sandbox root path not set. Use default: '%s'", config->downloadsFolder);
-            sprintf(dir, "%s/%s", config->downloadsFolder, id);
+            Log_Info("No parent workfolder. Using default: '%s'", config->downloadsFolder);
+            sprintf(dir, "%s", config->downloadsFolder);
```
JeffMill, line 1828: as unsafe.

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1804,57 +1804,86 @@ bool workflow_set_sandbox(ADUC_WorkflowHandle handle, const char* sandbox)
...
             ADUC_ConfigInfo_ReleaseInstance(config);
         }
         else
         {
             Log_Warn("Config is null. Use default sandbox root path: '%s'", DEFAULT_SANDBOX_ROOT_PATH);
-            sprintf(dir, "%s/%s", DEFAULT_SANDBOX_ROOT_PATH, id);
+            sprintf(dir, "%s", DEFAULT_SANDBOX_ROOT_PATH);
```
JeffMill, line 1834: do a safe strcpy

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1804,57 +1804,86 @@ bool workflow_set_sandbox(ADUC_WorkflowHandle handle, const char* sandbox)
...
         }
     }
 
-    free(pwf);
-    free(id);
+    if (dir[0] != 0)
```
JeffMill, line 1838: why would dir[0] be '\0' here?

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1804,57 +1804,86 @@ bool workflow_set_sandbox(ADUC_WorkflowHandle handle, const char* sandbox)
...
+
+    bdir = workflow_get_workflow_base_dir(handle);
+
+    if (bdir == NULL)
+    {
+        free(id);
+        return NULL;
+    }
+
+    sprintf(dir, "%s/%s", bdir, id);
```
JeffMill, line 1877: yikes, what's with these unchecked buffer copies?

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1804,57 +1804,86 @@ bool workflow_set_sandbox(ADUC_WorkflowHandle handle, const char* sandbox)
...
 
     if (dir[0] != 0)
```
JeffMill, line 1879: here too, how is dir[0] == '\0' here?

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1804,57 +1804,86 @@ bool workflow_set_sandbox(ADUC_WorkflowHandle handle, const char* sandbox)
...
+    {
+        return NULL;
+    }
+
+    // If workfolder explicitly specified, use it.
+    char* wf = workflow_get_string_property(handle, WORKFLOW_PROPERTY_FIELD_WORKFOLDER);
+    if (wf != NULL)
+    {
+        Log_Debug("Property '%s' not NULL - returning cached workfolder '%s'", WORKFLOW_PROPERTY_FIELD_WORKFOLDER, wf);
+        free(id);
```
JeffMill, line 1865: person who wrote this was against "goto done"?  A bunch of redundant free(id) calls.

JeffMill (changes requested):

src/utils/workflow_utils/src/workflow_utils.c
jw-msft: Consider ADUC_StringFormat as it seems a bit better, but note that it has a max size of 512 which is < PATH_MAX of 4096 in linux/limits.h so probably not viable option.

I feel it's better at this point to create a little util function in a .cpp, slap EXTERN_C_BEGIN/END around it in the .h, and use std::string += operator in the fn impl. The c-utility concat for STRING_HANDLE would be guaranteed to have more reallocations than c++ and the binary will already have the std::string methods from use elsewhere, so it's not adding more code to the elf.

An alternative, is something like this but it just feels like too much code to concatenate the string and a fwd slash:
if (pwf != NULL)
{
    memset(&dir, 0, sizeof(dir));
    const size_t DestLen = ARRAY_SIZE(dir) - 1; // -1 for the fwd-slash, and to be pedantic you can multiply by sizeof(char)
    const size_t SrcLen = ADUC_StrNLen(pwf, MAX_PATH); // I'm not really happy about this one--it could be better to have workflow_get_folder return the length too
    const size_t srcCharsCopied = ADUC_Safe_StrCopyN(&dir[0], pwf, DestByteLen, strlen(pwf));
    a util fn in .cpp, slap EXTERN_C_BEGIN/END around it in the .h, and use std::string += operator.
    bool truncated = srcCharsCopied < SrcLen;
    if (!truncated) {
        dir[srcCharsCopied] = '/';
        dir[srcCharsCopied + 1] = '\0';
    }
...
}

src/utils/workflow_utils/src/workflow_utils.c
jw-msft: FWIW, I added this to string_c_utils: size_t ADUC_Safe_StrCopyN(char* dest, const char* src, size_t destByteLen, size_t numSrcCharsToCopy).  If return val is < numSrcCharsToCopy, then it was truncated.

src/agent/adu_core_interface/src/adu_core_interface.c
nihemstr: Understood.

The workfolder is the workfolder for individual workflows.

In this case we're building the workflow folder before we launch into the workflow to do the root key json validation. Hence why we have this function here.

This file has been changed since the last review. Can you let me know if it still looks good? Otherwise I can explicitly change the naming convention to make the goals clearer.

src/inc/aduc/result.h
nihemstr: Result of running clang format. I need to address that bug.

src/utils/workflow_utils/src/workflow_utils.c
nihemstr: This should be addressed in a newer commit

src/utils/workflow_utils/src/workflow_utils.c
nihemstr: If there are nested workflows this allows the parent directory to be used for the processing of child workflows.

src/utils/workflow_utils/src/workflow_utils.c
nihemstr: Addressing this now.

src/agent/adu_core_interface/src/adu_core_interface.c
JeffMill: still confused.

Each workflow has a unique workfolder.  In that case,  `workflow_get_workfolder()` as was previously here seems fine to me.  Is there a reason you feel that `workflow_get_workflow_base_dir` is clearer?

src/libaducpal/inc/aducpal/limits.h
```diff
@@ -0,0 +1,22 @@
...
+#define ADUC_AL_LIMITS_H
+
+#ifdef ADUCPAL_USE_PAL
+
+#    ifdef __cplusplus
+extern "C"
+{
+#    endif
+
+#    define PATH_MAX 4096 // Maximum size of a filepath
```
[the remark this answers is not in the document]
JeffMill:
```
minwindef.h
60:#define MAX_PATH          260
```

src/utils/root_key_utils/src/root_key_util.c
```diff
@@ -805,32 +805,40 @@ ADUC_Result_t RootKeyUtility_GetReportingErc()
...
         goto done;
     }
 
-    storePackageJsonString = json_serialize_to_string(storePkgJsonValue);
-
-    if (storePackageJsonString == NULL)
+    if (localStore == NULL)
     {
-        goto done;
+        ADUC_Result temp = RootKeyUtility_ReloadPackageFromDisk(STRING_c_str(storePath), true);
```
JeffMill, line 827: what does "true" indicate? Add comment, e.g. `true /*reload_from_disk*/`

src/utils/root_key_utils/src/root_key_util.c
```diff
@@ -805,32 +805,40 @@ ADUC_Result_t RootKeyUtility_GetReportingErc()
...
+
+        if (IsAducResultCodeFailure(temp.ResultCode))
+        {
+            Log_Error("Failed to load RootKey Package from disk - recommending update to new package.");
```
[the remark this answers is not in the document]
JeffMill: oh, i see, "true" indicates an update required.  Maybe just "package load failed." and have caller trace that it's reloading?

Also, is there value in tracing "storePath" as well?

src/utils/root_key_utils/src/root_key_util.c
```diff
@@ -805,32 +805,40 @@ ADUC_Result_t RootKeyUtility_GetReportingErc()
...
+            return true;
```
JeffMill, line 832: nit: update_needed = true; goto done; for consistency

src/utils/rootkeypackage_utils/src/rootkeypackage_do_download.cpp
```diff
@@ -13,6 +13,8 @@
 
 EXTERN_C_BEGIN
 
+#define MAX_TIME_TO_DOWNLOAD_IN_HOURS 1 /* hour*/
```
JeffMill, line 16: I mentioned this previously and can't find my comment?  See the mail I sent on std::chrono.  C++ code shouldn't have `1 /* hour */`

JeffMill: tl;dr -- make this a strongly typed const with `= std::chrono::hours(1)`

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1804,6 +1816,107 @@ bool workflow_set_sandbox(ADUC_WorkflowHandle handle, const char* sandbox)
     return true;
 }
 
+bool _workflow_cpy_config_download_folder(char* dest, const size_t dest_size)
```
JeffMill, line 1819: what does `cpy` mean?

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1804,6 +1816,107 @@ bool workflow_set_sandbox(ADUC_WorkflowHandle handle, const char* sandbox)
...
+{
+    bool success = false;
+    const ADUC_ConfigInfo* config = ADUC_ConfigInfo_GetInstance();
+    if (config == NULL)
+    {
+        goto done; // could not get
+    }
+
+    size_t download_folder_len = ADUC_StrNLen(config->downloadsFolder, dest_size);
```
JeffMill, line 1828: you want strlen, not strnlen.  if you pass strnlen then you won't get larger than dest_size which means that you'll never get an error from safestrcopyn on 1836 which means you'll truncate the path which is **not** desirable.

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1804,6 +1816,107 @@ bool workflow_set_sandbox(ADUC_WorkflowHandle handle, const char* sandbox)
...
+
+    if (ADUC_Safe_StrCopyN(dest, config->downloadsFolder, dest_size, download_folder_len) != download_folder_len)
+    {
+        Log_Error("Failed copy config download folder");
+        goto done;
+    }
+
+    success = true;
+done:
+    ADUC_ConfigInfo_ReleaseInstance(config);
```
JeffMill, line 1844: ReleaseInstnace? Presumably GetInstance did locking or a refcount on the configinfo?

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1804,6 +1816,107 @@ bool workflow_set_sandbox(ADUC_WorkflowHandle handle, const char* sandbox)
...
+
+    return success;
+}
+
+// Returns the download base directory
+char* workflow_get_workflow_base_dir(const ADUC_WorkflowHandle handle)
```
JeffMill, line 1850: seems like overkill to me.  Dont we have ADUC_StringFormat that does allocation?  If so, just call ADUC_StringFormat("%s/%s", workfolder, path); or whatever and return that.

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -2507,6 +2619,8 @@ workflow_init_from_file(const char* updateManifestFile, bool validateManifest, A
         goto done;
     }
 
+    // address of rootJsonValue = 0x5555558 debb0
```
JeffMill, line 2622: remove 2622-2623.

JeffMill (dismissed):
approved with comments.

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1258,6 +1266,7 @@ ADUC_Result _workflow_parse(JSON_Value* updateActionJson, bool validateManifest,
     // commit ownership of cloned JSON_Value to the workflow's UpdateActionObject.
     updateActionObject = json_value_get_object(updateActionJsonClone);
```
jw-msft, line 1267: Could we just set wf->UpdateActionObject here directly and forgo the updateActionObject local since it's just getting set to NULL immediately anyways and never getting used, or is it to make it clearer?

JeffMill (approved):
Approved with suggestions.

src/extensions/download_handlers/plugin_examples/microsoft_delta_download_handler/source_update_cache/src/source_update_cache_utils.c
```diff
@@ -109,13 +109,13 @@ STRING_HANDLE ADUC_SourceUpdateCacheUtils_CreateSourceUpdateCachePath(
     // file path format:
     //     {base_dir}/{provider}/{hashAlg}-{hash}
 
-    sanitizedProvider = SanitizePathSegment(provider);
+    sanitizedProvider = PathUtils_SanitizePathSegment(provider);
```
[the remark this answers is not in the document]
jw-msft: I think for the C utils we try to prefix them with name of the CU as a cash-strapped namespace, but it is inconsistently applied--I've open a tracking issue internally to make the naming convention align.

src/utils/path_utils/src/path_utils.c
```diff
@@ -11,13 +11,14 @@
...
 
+#include <aducpal/limits.h> // for PATH_MAX
 /**
  * @brief Replaces non-alphunumeric chars with _ (underscore) char for use in path segments of a file path.
  * @param unsanitized The string to be sanitized
  * @return STRING_HANDLE The file path string, or NULL on error.
  * @details Caller owns it and must call STRING_delete() when done with it.
  */
-STRING_HANDLE SanitizePathSegment(const char* unsanitized)
+STRING_HANDLE PathUtils_SanitizePathSegment(const char* unsanitized)
```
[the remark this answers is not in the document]
jw-msft: it's definitely not going to if the codepoint is one that encodes to 2 or 3 bytes for the utf-8 encoding. Should certainly track an issue on it and determine if it is a ship-blocker, but I think there are likely some other util fn that assume ascii charset for paths so an audit would be needed to fix it holistically. //cc @eshashah-msft

jw-msft (dismissed):

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1813,40 +1813,40 @@ bool workflow_set_sandbox(ADUC_WorkflowHandle handle, const char* sandbox)
...
     {
         Log_Error("Invalid base sandbox dir: '%s'", config->downloadsFolder);
         goto done;
     }
 
-    if (ADUC_Safe_StrCopyN(dest, config->downloadsFolder, dest_size, download_folder_len) != download_folder_len)
+    if (!mallocAndStrcpy_s(&downloads_folder_path, config->downloadsFolder))
     {
-        Log_Error("Failed copy config download folder");
+        Log_Error("Failed to allocate memory for downloads folder path");
```
jw-msft, line 1836: Should probably omit this as the logging would likely fail in this case too unless logging optimistically?

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1855,69 +1855,61 @@ char* workflow_get_root_sandbox_dir(const ADUC_WorkflowHandle handle)
...
             goto done;
         }
     }
 
-    if (mallocAndStrcpy_s(&tempRet, dir) != 0)
+    if (mallocAndStrcpy_s(&ret, tempRet) != 0)
     {
-        Log_Error("Failed to allocate memory for workflow base dir");
-        tempRet = NULL;
+        Log_Error("Failed allocate the root sandbox dir");
```
jw-msft, line 1888: omit trace

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1929,8 +1921,17 @@ char* workflow_get_workfolder(const ADUC_WorkflowHandle handle)
...
+    id = workflow_get_id(handle);
+    size_t id_len = ADUC_StrNLen(id, WORKFLOW_DIR_PATH_MAX_SIZE);
+
+    if (id_len == 0 || id_len == WORKFLOW_DIR_PATH_MAX_SIZE)
+    {
+        Log_Error("Workflow id is too long to be in a path: '%s'", id);
+        goto done;
+    }
+
+    ret = PathUtils_ConcatenateDirAndFolderPaths(base_sandbox_dir, id);
```
jw-msft, line 1933: Since base_sandbox_dir has type `char*`, shouldn't it be passing &base_sandbox_dir so it can set the contents of it with newly allocated str?

src/rootkey_workflow/src/rootkey_workflow.c
```diff
@@ -100,19 +100,19 @@ ADUC_Result RootKeyWorkflow_UpdateRootKeys(const char* workflowId, const char* w
 
 #ifndef ADUC_ENABLE_SRVC_E2E_TESTING
 
-#ifdef ADUC_E2E_TESTING_ENABLED
+#    ifdef ADUC_E2E_TESTING_ENABLED
```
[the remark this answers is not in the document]
Jerry-GeGarcia: agree with this point. Even if it is not addressed immediately as I imagine it would be a more involved change.

Jerry-GeGarcia (dismissed):
:shipit:

JeffMill (dismissed):
approved with comments

src/utils/path_utils/src/path_utils.c
```diff
@@ -59,3 +60,82 @@ STRING_HANDLE SanitizePathSegment(const char* unsanitized)
...
+ * @brief Concatenates a directory and folder to form a path.
+ * @param dirPath The directory path (eg /var/lib/adu)
+ * @param folderName The folder name (eg 12345678-1234-1234-1234-123456789012)
+ * @return char* The concatenated path, or NULL on error.
+*/
+char* PathUtils_ConcatenateDirAndFolderPaths(const char* dirPath, const char* folderName)
+{
+    char* ret = NULL;
+    char* tempRet = NULL;
+    char dir[PATH_MAX + 1] = { 0 }; // Plus one for null terminator
```
jw-msft, line 74: PATH_MAX already includes for the null terminator, so should remove + 1 here: https://www.gnu.org/software/libc/manual/html_node/Limits-for-Files.html

src/utils/path_utils/src/path_utils.c
```diff
@@ -59,3 +60,82 @@ STRING_HANDLE SanitizePathSegment(const char* unsanitized)
...
+    const char* file_delimeter = "/";
+    size_t file_delimeter_len = strlen(file_delimeter); /* can use strlen since this is a known good string */
+
+    if (IsNullOrEmpty(dirPath) || IsNullOrEmpty(folderName))
+    {
+        goto done;
+    }
+
+    size_t dirPathLen = ADUC_StrNLen(dirPath, PATH_MAX);
```
jw-msft, line 83: `PATH_MAX-1` and same for below, or better yet `ARRAY_SIZE(dir)`

jw-msft (changes requested):

src/utils/path_utils/src/path_utils.c
```diff
@@ -59,3 +60,82 @@ STRING_HANDLE SanitizePathSegment(const char* unsanitized)
...
+ * @param folderName The folder name (eg 12345678-1234-1234-1234-123456789012)
+ * @return char* The concatenated path, or NULL on error.
+*/
+char* PathUtils_ConcatenateDirAndFolderPaths(const char* dirPath, const char* folderName)
+{
+    char* ret = NULL;
+    char* tempRet = NULL;
+    char dir[PATH_MAX + 1] = { 0 }; // Plus one for null terminator
+    const char* file_delimeter = "/";
+    size_t file_delimeter_len = strlen(file_delimeter); /* can use strlen since this is a known good string */
```
jw-msft, line 76: nit, no need to call strlen: const size_t file_delimeter_len = 1;

src/utils/path_utils/src/path_utils.c
```diff
@@ -59,3 +60,82 @@ STRING_HANDLE SanitizePathSegment(const char* unsanitized)
...
     return sanitizedHandle;
 }
+
+/**
+ * @brief Concatenates a directory and folder to form a path.
+ * @param dirPath The directory path (eg /var/lib/adu)
+ * @param folderName The folder name (eg 12345678-1234-1234-1234-123456789012)
+ * @return char* The concatenated path, or NULL on error.
+*/
+char* PathUtils_ConcatenateDirAndFolderPaths(const char* dirPath, const char* folderName)
```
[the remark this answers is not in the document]
jw-msft: JeffMill's suggestion here is looking simpler than current impl. Why not do it this way?

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1804,57 +1813,133 @@ bool workflow_set_sandbox(ADUC_WorkflowHandle handle, const char* sandbox)
...
-        sprintf(dir, "%s/%s", pwf, id);
 
-        Log_Debug("Using parent workfolder: '%s/%s'", pwf, id);
+        if (pwf == NULL)
+        {
+            Log_Error("Failed to get parent workfolder");
+            goto done;
+        }
+
+        size_t pwf_len = ADUC_StrNLen(pwf, WORKFLOW_DIR_PATH_MAX_SIZE);
```
jw-msft, line 1864: Does WORKFLOW_DIR_PATH_MAX_SIZE include nul-term? If so, then needs to be -1 here.

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1804,57 +1813,133 @@ bool workflow_set_sandbox(ADUC_WorkflowHandle handle, const char* sandbox)
...
+    // Return ([parent's workfolder] or [default sandbox folder]) + "/" + [workflow id];
+    base_sandbox_dir = workflow_get_root_sandbox_dir(handle);
+
+    if (base_sandbox_dir == NULL)
+    {
+        goto done;
+    }
+
+    id = workflow_get_id(handle);
+    size_t id_len = ADUC_StrNLen(id, WORKFLOW_DIR_PATH_MAX_SIZE);
```
jw-msft, line 1923: same here. likely off by one err

jw-msft (approved):

Nox-MSFT (approved):

PR #584, opened 2023-12-05
Add scoped topics, suback response topic before publish

jw-msft:
- Distinguishes between scoped topics (operations other than enrollment) and non-scoped topics (enrollment response will contain the scope suffix for later operations)
- Matches correlation id from the mqtt correlation-data to pair the subscripbe with its acknowledgement from the mqtt broker
- various fixes and tweaks
- deletion of some unused/redundant codec
- made state store in-memory only to simplify (will add persistence back later)


Jerry-GeGarcia:
>  *  For 'ainfo_resp' messages, if the Correlation Data matches the 'en,the client should parse the message and update the agentinfo state.

what is 'en?

---
Refers to: src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/adu_agent_info_management.c:65 in 4720e51. [](commit_id = 4720e512a20cf4eb0d01d8f7f504bbe837449c27, deletion_comment = False)

src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/adu_agent_info_management.c
```diff
@@ -143,20 +108,61 @@ void OnMessage_ainfo_resp(
...
+            retriable_operation_context->retryFunc(retriable_operation_context, retriable_operation_context->retryParams);
+        }
+        break;
+
+    case MQTT_RC_TOPIC_NAME_INVALID: // fall-through
+    case MQTT_RC_PACKET_ID_IN_USE: // fall-through
+    case MQTT_RC_PACKET_TOO_LARGE: // fall-through
+    case MQTT_RC_QUOTA_EXCEEDED:
+        if (retriable_operation_context != NULL)
+        {
```
Jerry-GeGarcia, line 159: trace worthy?

src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/adu_enrollment_management.c
```diff
@@ -152,22 +107,65 @@ void OnMessage_enr_resp(
...
+        Log_Error("Fail handling enrollment response.");
         goto done;
     }
 
 done:
     free(scopeId);
+    free(correlationData);
+}
+
+void OnPublish_enr_resp(struct mosquitto* mosq, void* obj, const mosquitto_property* props, int reason_code)
```
Jerry-GeGarcia, line 129: >OnPublish_enr_resp [](http://example.com/codeflow?start=5&length=18)

this is almost a copy of OnPublish_ainfo_resp; some refactoring in order?

src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/agentinfo_request_operation.c
```diff
@@ -195,7 +219,8 @@ static bool HandlingRequestAgentInfo(ADUC_Retriable_Operation_Context* context,
         ADUC_Retriable_Set_State(context, ADUC_Retriable_Operation_State_InProgress);
 
         // is the current request timed-out?
-        if (context->lastExecutionTime + context->operationTimeoutSecs < nowTime)
+        // TODO: use context->operationTimeoutSecs instead of 30 once operationTimeoutSecs set correctly to non-zero.
```
Jerry-GeGarcia, line 222: > operationTimeoutSecs set correctly to non-zero [](http://example.com/codeflow?start=69&length=47)

server side?

src/extensions/agent_modules/shared/common/inc/aduc/adu_module_state.h
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * @brief The module state.
+ */
+typedef struct tagADUC_MQTT_CLIENT_MODULE_STATE
+{
+    void* moduleInitData; //!< Module init data
+    ADUC_MQTT_SETTINGS mqttSettings; //!< MQTT connection settings
+
+    char* mqtt_topic_service2agent; // For v1, we're only subscribe to 1 topic.
+    char* mqtt_topic_agent2service; // For v1, we're only publish to 1 topic.
```
Jerry-GeGarcia, line 19: >we're [](http://example.com/codeflow?start=47&length=5)

we only

src/extensions/agent_modules/shared/common/inc/aduc/adu_module_state.h
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * @brief The module state.
+ */
+typedef struct tagADUC_MQTT_CLIENT_MODULE_STATE
+{
+    void* moduleInitData; //!< Module init data
+    ADUC_MQTT_SETTINGS mqttSettings; //!< MQTT connection settings
+
+    char* mqtt_topic_service2agent; // For v1, we're only subscribe to 1 topic.
```
Jerry-GeGarcia, line 18: >subscribe [](http://example.com/codeflow?start=58&length=9)

subscribed

Jerry-GeGarcia:
>  * @brief Implements the Device Update agent state store functionality.

I could use another pair of eyes on this

---
Refers to: src/extensions/agent_modules/shared/state_store/src/agent_state_store.c:3 in 4720e51. [](commit_id = 4720e512a20cf4eb0d01d8f7f504bbe837449c27, deletion_comment = False)

Jerry-GeGarcia (approved):
:shipit:

src/extensions/agent_modules/shared/state_store/src/agent_state_store.c
```diff
@@ -62,379 +68,65 @@ static char* SafeStrdup(const char* str)
...
-    }
+        Log_Info("State store terminated successfully.");
+        isInitialized = false;
 
-    *out_value = SafeStrdup(result); // Callers are responsible for freeing this memory.
-    if (*out_value == NULL)
-    {
-        Log_Error("Failed to duplicate string value for key %s.", key);
-        goto done;
+        sem_destroy(&state_semaphore);
```
nihemstr, line 121: Do we use this in counting or binary? Does this protect the actual state variable? We should acquire the semaphore at the beginning and have a way to block any operations after assuming this is meant for shared memory.

src/extensions/agent_modules/shared/state_store/src/agent_state_store.c
```diff
@@ -62,379 +68,65 @@ static char* SafeStrdup(const char* str)
...
-    JSON_Object* root_object = GetRootObject(durable);
-    int result = json_object_dotget_boolean(root_object, key);
-
-    if (json_object_dotget_value(root_object, key) == NULL
-        || json_value_get_type(json_object_dotget_value(root_object, key)) != JSONBoolean)
+    else
     {
-        Log_Warn("Failed to retrieve bool value for key %s.", key);
-        goto done;
+        Log_Info("Nothing to deinitialize.");
```
nihemstr, line 125: I don't think we need to log when this is called. I'm not sure how we'd use this to determine any issue given we're either uninitialized (which might be a good debug trace at the beginning of the function) or we can see the trace where we've already uninited the state before and we're suddenly trying again?

src/extensions/agent_modules/shared/state_store/src/agent_state_store.c
```diff
@@ -62,379 +68,65 @@ static char* SafeStrdup(const char* str)
...
-    return res;
 }
 
-ADUC_STATE_STORE_RESULT ADUC_StateStore_SetBool(bool durable, const char* key, int value)
-{
-    return ADUC_StateStore_SetData(durable, key, json_value_init_boolean(value));
-}
-
-const char* ADUC_StateStore_GetExternalDeviceId(void)
+const char* ADUC_StateStore_GetExternalDeviceId()
```
nihemstr, line 129: We should do a copy of the value here and not take a constant pointer to something that can be changed in other sections of the code outside of the semaphore.

src/extensions/agent_modules/shared/state_store/src/agent_state_store.c
```diff
@@ -496,10 +188,12 @@ int ADUC_StateStore_GetDeviceRegistrationStatePollIntervalSeconds()
  * @brief Get the Device Update service device ID.
  * @return The device ID, or NULL if not found.
  */
-const char* ADUC_StateStore_GetDeviceId(void)
+const char* ADUC_StateStore_GetDeviceId()
```
nihemstr, line 191: All of these getters should return copies of the value instead of the actual value or the callers need to lock the semaphore and manage access themselves.

src/extensions/agent_modules/shared/state_store/src/agent_state_store.c
```diff
@@ -513,251 +207,244 @@ ADUC_STATE_STORE_RESULT ADUC_StateStore_SetDeviceId(const char* deviceId)
...
+ADUC_STATE_STORE_RESULT ADUC_StateStore_SetScopeId(const char* scopeId)
 {
+    ADUC_STATE_STORE_RESULT result = ADUC_STATE_STORE_RESULT_ERROR;
+
     sem_wait(&state_semaphore);
-    free(state.deviceUpdateServiceInstance);
-    state.deviceUpdateServiceInstance = SafeStrdup(instanceName);
+
+    free(state.scopeId);
+    state.scopeId = SafeStrdup(scopeId);
```
nihemstr, line 246: maybe nit: do we know these will always be wellformed strings? Do we need to do ADUC_StrCopyN()

src/extensions/agent_modules/shared/state_store/src/agent_state_store.c
```diff
@@ -513,251 +207,244 @@ ADUC_STATE_STORE_RESULT ADUC_StateStore_SetDeviceId(const char* deviceId)
...
+ * @param topic The topic.
+ * @param isScoped True to set the subscribed status of the single scoped topic. False to set the subscribed status of the single non-scoped topic.
+ * @param newStatus The new subscribe status for the topic.
+ * @return ADUC_STATE_STORE_RESULT_OK on successful setting of subscribe/unsubscribe status.
+ * @return ADUC_STATE_STORE_RESULT_ERROR_EMPTY_TOPIC if the topic is an empty or null string.
+ * @return ADUC_STATE_STORE_RESULT_ERROR_MAX_TOPIC_BYTE_LENGTH_EXCEEDED if the utf-8 topic string's byte length exceeds MAX_ADU_MQTT_TOPIC_BYTE_LEN
+ * @return ADUC_STATE_STORE_RESULT_UNKNOWN_TOPIC when trying to unsubscribe from the scoped or non-scoped topic that is not currently subscribed.
+ * @return ADUC_STATE_STORE_RESULT_ERROR when failed subscribe/unsubscribe an existing topic for some other reason.
+ */
+ADUC_STATE_STORE_RESULT ADUC_StateStore_SetTopicSubscribedStatus(const char* topic, bool isScoped, bool subscribed)
```
nihemstr, line 299: is there a call to this function that would never have subscribed set to true?

src/extensions/agent_modules/shared/state_store/tests/agent_state_store_ut.cpp
```diff
@@ -20,14 +20,14 @@ TEST_CASE("communication handle")
 {
     SECTION("Valid handle")
     {
-        ADUC_STATE_STORE_RESULT result = ADUC_StateStore_Initialize(TEST_STATE_FILE);
+        ADUC_STATE_STORE_RESULT result = ADUC_StateStore_Initialize(TEST_STATE_FILE, false /* isUsingProvisioningService */);
```
nihemstr, line 23: May not be apart of this PR but we should start being more aggressive about happy + unhappy paths making sure we're exiting the way we want.

src/utils/c_utils/src/string_c_utils.c
```diff
@@ -438,3 +438,43 @@ size_t ADUC_Safe_StrCopyN(char* dest, const char* src, size_t destByteLen, size_
...
+* @return 0 on success, -2 on invalid argument, and -1 for other errors.
+*/
+int ADUC_AllocAndStrCopyN(char** dest, const char* src, size_t srcLen)
+{
+    int res = -1;
+
+    size_t num_copied = -1;
+    char* tmpDest = NULL;
+
+    if (dest == NULL || srcLen == 0)
```
nihemstr, line 456: Add IsNullOrEmpty(src)

nihemstr (changes requested):
Added some comments. Not really sure how to approach reading the overall flow but was able to get through some of the bigger stuff in this PR.

The ADUC_Store getters and setters running around the semaphores needs to be changed for the operations to prevent misreads / miswrites (eg someone is setting the deviceId while another person is reading the const char* value returned by the function).

src/extensions/agent_modules/cloud_communication/communication_channel/mqtt_broker/src/adu_communication_channel.c
```diff
@@ -800,43 +794,55 @@ void ADUC_Communication_Channel_OnSubscribe(
...
 
         if (commMgrState->mqttCallbacks.on_subscribe_v5 != NULL)
         {
             commMgrState->mqttCallbacks.on_subscribe_v5(
                 mosq, commMgrState->ownerModuleContext, mid, qos_count, granted_qos, props);
         }
     }
+    else
+    {
+        Log_Debug("No pending sub found with mid: %d", mid);
```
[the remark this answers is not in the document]
jw-msft: I changed these to messageId in the latest PR highest in the stack

src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/inc/aduc/adu_agent_info_management.h
```diff
@@ -18,27 +18,9 @@ EXTERN_C_BEGIN
...
-/**
- * @brief Callback should be called when the client receives an agent info response message from the Device Update service.
- * @param mosq The mosquitto instance making the callback.
- * @param obj The user data provided in mosquitto_new. This is the module state
- * @param msg The message data.
- * @param props The MQTT v5 properties returned with the message.
- */
-void OnMessage_ainfo_resp(
-    struct mosquitto* mosq, void* obj, const struct mosquitto_message* msg, const mosquitto_property* props);
+void OnMessage_ainfo_resp(struct mosquitto* mosq, void* obj, const struct mosquitto_message* msg, const mosquitto_property* props);
```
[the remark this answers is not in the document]
jw-msft: I'll do a separate PR to touch and clang-format everything before merging into feature branch. I didn't want to generate lots of clang-format noise in these stacked ones because all the new files basically had needed it.

jw-msft:
> > * For 'ainfo_resp' messages, if the Correlation Data matches the 'en,the client should parse the message and update the agentinfo state.
> 
> what is 'en? Oh, enrollment.
> 
> Refers to: src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/adu_agent_info_management.c:65 in [4720e51](https://github.com/Azure/iot-hub-device-update/commit/4720e512a20cf4eb0d01d8f7f504bbe837449c27). [](commit_id = [4720e51](https://github.com/Azure/iot-hub-device-update/commit/4720e512a20cf4eb0d01d8f7f504bbe837449c27), deletion_comment = False)

No actually it's a serious typo--I think I was trying to say `matches, then the client` but I was typing way to fast with probably some editor key-seq macros mixed in.

src/extensions/agent_modules/cloud_communication/communication_channel/mqtt_broker/src/adu_communication_channel.c
```diff
@@ -800,43 +794,55 @@ void ADUC_Communication_Channel_OnSubscribe(
...
             break;
         }
     }
 
     if (subInfo != NULL)
     {
-        Log_Debug("move pending sub -> subscribed");
+        Log_Debug("move pending topic '%s' -> subscribed", topic);
+
+        ADUC_SetCommunicationChannelState(commMgrState, ADU_COMMUNICATION_CHANNEL_CONNECTION_STATE_SUBSCRIBED);
```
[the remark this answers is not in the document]
jw-msft: It is currently thread-safe as there's only one thread as it stems from mosquitto_loop() call after DoWork call from main() that would result in callbacks like OnConnect, OnPublish, OnMessage, etc.  I am adding another thread to handle items on the workqueue in an upcoming PR but the additional thread will update state store with update result reporting json and then the main thread will pick that up and interact with libmosquitto to send the updrslt_req, handle puback and adu service updrslt_resp response.

src/extensions/agent_modules/shared/state_store/src/agent_state_store.c
jw-msft: Good point. Not currently because in ADUC_Communication_Channel_OnSubscribe() it ignores the reason code--if it's a failure reason code, then it should call this with false so that it's not subscribed because the SUBACK had a failure.

Opened this bug on it to track fixing it:

[Bug 48162686](https://microsoft.visualstudio.com/OS/_workitems/edit/48162686): [Gen2] ADUC_Communication_Channel_OnSubscribe needs to check mqtt reason code of SUBACK to know when topic subscribe failed

src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/adu_agent_info_management.c
jw-msft: yes. Added with this commit: [add tracing for puback failure and clang-format](https://github.com/Azure/iot-hub-device-update/pull/584/commits/ba38bbca78bf548ccec83aa3ef9cd0cc93b6c9e1)

src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/adu_enrollment_management.c
jw-msft: Factored out into adu_mqtt_common.c now with this commit: [factor out puback handling to adu_mqtt_common](https://github.com/Azure/iot-hub-device-update/pull/584/commits/bd1513cbd9eadca3c66734cad1cfdf88fe227371)

src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/agentinfo_request_operation.c
jw-msft: It is currently a default setting in the client side, but we are having conversations about also having poll freq and retry/timeout settings flow down from service.

src/extensions/agent_modules/shared/common/inc/aduc/adu_module_state.h
jw-msft: Fixed with this commit: [fix comment - we only](https://github.com/Azure/iot-hub-device-update/pull/584/commits/a58ea8c43d782d7df92582bb216e752b26326c8f)

src/extensions/agent_modules/shared/common/inc/aduc/adu_module_state.h
jw-msft: went with imperative (we only subscribe)

src/extensions/agent_modules/cloud_communication/communication_channel/mqtt_broker/src/adu_communication_channel.c
```diff
@@ -800,43 +794,55 @@ void ADUC_Communication_Channel_OnSubscribe(
...
                 mosq, commMgrState->ownerModuleContext, mid, qos_count, granted_qos, props);
         }
     }
+    else
+    {
+        Log_Debug("No pending sub found with mid: %d", mid);
+    }
 }
 
+#define LOG_ALL_MOSQUITTO 1
```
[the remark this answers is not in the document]
jw-msft: done with commit: [move LOG_ALL_MOSQUITTO define to top](https://github.com/Azure/iot-hub-device-update/pull/584/commits/e52c96e7eaaecc06290575209cf4809ea9e9f3a1)

src/extensions/agent_modules/cloud_communication/communication_channel/mqtt_broker/src/adu_communication_channel.c
```diff
@@ -1339,77 +1348,23 @@ int ADUC_Communication_Channel_MQTT_Subscribe(
          goto done;
     }
 
-    // Release ownership of fields in the callbackInfo Element as those were wholesale copied into the VECTOR_HANDLE's block
-    // When element is erased from vector, the fields will need freeing at that time.
-    memset(&callbackInfo, 0, sizeof(callbackInfo));
+    ADUC_SetCommunicationChannelState(commMgrState, ADU_COMMUNICATION_CHANNEL_CONNECTION_STATE_SUBSCRIBING);
+    mosqResult = MOSQ_ERR_SUCCESS;
```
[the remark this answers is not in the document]
jw-msft: Resolved with commit: [move set of mosqResult success right before done label](https://github.com/Azure/iot-hub-device-update/pull/584/commits/dcbb183f522fb604b3083a20b681ba726d266eef)

src/extensions/agent_modules/cloud_communication/communication_channel/mqtt_broker/src/adu_communication_channel.c
```diff
@@ -1339,77 +1348,23 @@ int ADUC_Communication_Channel_MQTT_Subscribe(
...
-        *mid = callbackInfo.messageId;
-    }
+    *outMessageId = callbackInfo.messageId;
 
-    Log_Debug("-> SUB %s mid[%d]", topic, callbackInfo.messageId);
-    ADUC_SetCommunicationChannelState(commMgrState, ADU_COMMUNICATION_CHANNEL_CONNECTION_STATE_SUBSCRIBING);
-    mosqResult = MOSQ_ERR_SUCCESS;
+    // Release ownership of fields in the callbackInfo Element.
+    // When element is erased from vector, the fields will need freeing at that time.
+    memset(&callbackInfo, 0, sizeof(callbackInfo));
```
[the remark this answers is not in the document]
jw-msft: as discussed offline, VECTOR_push_back copies the item's struct memory so effectively is a shallow copy of fields and, thus, transfers ownership of ptr fields.

src/extensions/agent_modules/shared/state_store/src/agent_state_store.c
jw-msft: good catch. the pointers need an alloc and deep copy in general. Fixed with this commit:[make copies of strings for state store getters](https://github.com/Azure/iot-hub-device-update/pull/584/commits/8635ed011ebfc7396ecc82ffe15532439285e6df)

src/extensions/agent_modules/shared/state_store/src/agent_state_store.c
jw-msft: Good catch. Now doing a ADUC_StrNLen check and returning ADUC_STATE_STORE_RESULT_ERROR if == 0 or == MAX_ADU_SCOPE_ID_BYTE_LEN. Also, copying with MallocAndSubstr for now since it has N bound and not as churned as ADUC_StrCopyN with later PRs (less conflicts for now).

Here is the commit that addresses this: [bound state store set of scopeId by max scopeId bytelen](https://github.com/Azure/iot-hub-device-update/pull/584/commits/720d02ee1e2ef727b0e93efa3cd56824667303f8)

src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/adu_agent_info_management.c
```diff
@@ -119,15 +75,24 @@ void ADUC_AgentInfo_Management_Destroy(ADUC_AGENT_MODULE_HANDLE handle)
...
+    if (retriableOperationContext == NULL || agentInfoData == NULL || messageContext == NULL)
+    {
+        Log_Error("Invalid callback obj");
+        goto done;
+    }
 
     json_print_properties(props);
 
-    if (!ADU_are_correlation_ids_matching(props, messageContext->correlationId))
+    if (!ADU_are_correlation_ids_matching(props, messageContext->correlationId, &correlationData, &correlationDataByteLen))
```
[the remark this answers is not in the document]
jw-msft: Fixed with following commit: [use size_t in fn sig in lieu of uint16_t](https://github.com/Azure/iot-hub-device-update/pull/584/commits/fce31ee12f3c577aa9df76feac82638636ac45a1)

src/extensions/agent_modules/shared/state_store/src/agent_state_store.c
jw-msft: addressed with commit: [make copies of strings for state store getters](https://github.com/Azure/iot-hub-device-update/pull/584/commits/8635ed011ebfc7396ecc82ffe15532439285e6df)

src/extensions/agent_modules/shared/state_store/src/agent_state_store.c
jw-msft: removed with commit: [omit trace when called and already deinit'ed](https://github.com/Azure/iot-hub-device-update/pull/584/commits/5bf0362a8871f2cd51c3aae2e86c9fc4ff8fa46c)

src/extensions/agent_modules/shared/state_store/src/agent_state_store.c
jw-msft: By design. No need to acquire the semaphore becaue this will only be called after we exit the loop that was calling the doWork on the agent modules so nothing and also having stopped the worker threads, so it would not be competing with any other thread. For the same reason, no need to block operations after. It should probably just be a mutex and will change it to that later.

src/extensions/agent_modules/shared/state_store/src/agent_state_store.c
```diff
@@ -62,379 +68,65 @@ static char* SafeStrdup(const char* str)
 /**
  * @brief Initialize the state store.
  * @param state_file_path The path to the state file.
+ * @param usingProvisioningService true if using service to provision device; otherwise, using config-based provisioning.
  * @return ADUC_STATE_STORE_RESULT_OK on success, ADUC_STATE_STORE_RESULT_ERROR on failure.
+ * @remark The assumption is that only a single thread is every calling this.
  */
-ADUC_STATE_STORE_RESULT ADUC_StateStore_Initialize(const char* state_file_path)
+ADUC_STATE_STORE_RESULT ADUC_StateStore_Initialize(const char* state_file_path, bool usingProvisioningService)
```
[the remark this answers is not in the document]
jw-msft: Yes, I believe that is removed in a later PR in the stack.

src/extensions/agent_modules/shared/state_store/src/agent_state_store.c
```diff
@@ -38,21 +40,25 @@ static sem_t state_semaphore;
...
     bool isAgentInfoReported;
+    bool isDeviceProvisionedByService;
 } ADUC_StateData;
 
 static ADUC_StateData state = { 0 }; // Initialize to zero.
 
 static char* SafeStrdup(const char* str)
 {
-    if (str == NULL)
+    if (IsNullOrEmpty(str))
```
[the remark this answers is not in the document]
jw-msft: I plan on removing this if haven't already in later PR in the stack.

src/extensions/agent_modules/shared/state_store/src/agent_state_store.c
```diff
@@ -8,19 +8,21 @@
...
+#define ADUC_DEFAULT_DEVICE_REGISTRATION_STATE_POLL_INTERVAL_SECONDS (10)
 
+// Internal state data
 static bool isInitialized = false;
 
+// MQTT SPEC says topics can have max len of 65536 utf-8 encoded bytes.
+//
+// TODO: Figure out a lower, tighter maximum based on topic template path segment maximums, or if this maximum is not sufficient.
+#define MAX_ADU_MQTT_TOPIC_BYTE_LEN 2048
+
```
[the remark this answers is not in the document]
jw-msft: The technical max number of utf-8 bytes for topics (according to MQTT spec) is 65536 bytes, so I've increased it to that so that we do not artificially preclude future scenarios.

I've also updated `MAX_ADU_DEVICE_ID_BYTE_LEN` and `MAX_ADU_SCOPE_ID_BYTE_LEN` based on tracking those down.

src/extensions/agent_modules/shared/mosquitto_utils/src/adu_mosquitto_utils.c
```diff
@@ -73,28 +73,42 @@ bool ADU_mosquitto_get_correlation_data(const mosquitto_property* props, char**
...
-        mosquitto_property_read_binary(props, MQTT_PROP_CORRELATION_DATA, (void**)&cid, &cidLen, false);
+    const mosquitto_property* p = mosquitto_property_read_binary(props, MQTT_PROP_CORRELATION_DATA, (void**)&cid, &cidLen, false);
     if (p == NULL || cid == NULL)
     {
         goto done;
     }
 
+    if (outCorrelationData != NULL)
+    {
+        if (ADUC_AllocAndStrCopyN(outCorrelationData, cid, cidLen) != 0)
```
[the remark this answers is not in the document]
jw-msft: Thanks. Fixed with commit: [cast uint16_t to size_t for ADUC_AllocAndStrCopyN](https://github.com/Azure/iot-hub-device-update/pull/584/commits/a70cf7d0ef19ae5d40e0f80d469899f339af411c)

src/extensions/agent_modules/shared/mosquitto_utils/src/adu_mosquitto_utils.c
```diff
@@ -73,28 +73,42 @@ bool ADU_mosquitto_get_correlation_data(const mosquitto_property* props, char**
...
-    if (props == NULL)
+    if (props == NULL || IsNullOrEmpty(correlationId))
     {
         goto done;
     }
 
     // Check if correlation data matches the latest enrollment message.
-    const mosquitto_property* p =
-        mosquitto_property_read_binary(props, MQTT_PROP_CORRELATION_DATA, (void**)&cid, &cidLen, false);
+    const mosquitto_property* p = mosquitto_property_read_binary(props, MQTT_PROP_CORRELATION_DATA, (void**)&cid, &cidLen, false);
```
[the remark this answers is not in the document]
jw-msft: not magic, just mosquitto_property_read_binary requires void** for this argument and the compiler won't allow passing a char** to it and the char is 8 bits so perfect for bytes of binary data. Under the hood it uses calloc and memcpy that both return and take void* so I suppose the API author didn't want to mess with the void* typing by casting things to char*.

I've summarized all this in a comment in this commit: [add comment about void** passed to mosquitto_property_read_property](https://github.com/Azure/iot-hub-device-update/pull/584/commits/8deef67a05e0751d1e7e81ca476e1f8ad067ab94)

src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/adu_agent_info_management.c
```diff
@@ -143,20 +108,61 @@ void OnMessage_ainfo_resp(
         goto done;
     }
 
-    // NOTE: Do no parse ainfo_resp msg payload as it is empty.
+    // NOTE: Do no parse ainfo_resp msg payload as it is empty as per client/server protocol design.
```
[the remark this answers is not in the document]
jw-msft: done with this commit: [prefix ParseAndValidateCommon.. with ADU_MosquittoUtils_](https://github.com/Azure/iot-hub-device-update/pull/584/commits/bb3b2c9e383424e229ecfa1d32fdb45d1c4a7fb7)

Opened this tracking bug to prefix the others: [Bug 48195268](https://microsoft.visualstudio.com/OS/_workitems/edit/48195268): [Gen2] Make Consistent ADU_MosquittoUtils_ Prefix for all functions in adu_mosquitto_utils

src/extensions/agent_modules/shared/mosquitto_utils/src/adu_mosquitto_utils.c
```diff
@@ -73,28 +73,42 @@ bool ADU_mosquitto_get_correlation_data(const mosquitto_property* props, char**
  * @brief Check if the correlation data in the MQTT properties matches the provided correlation ID.
  * @param[in] props Pointer to the MQTT properties from which to retrieve the correlation data.
  * @param[in] correlationId The correlation ID to check against.
+ * @param[out] outCorrelationData Optional. If not NULL, will set to a string with the value of mqtt correlation-data property.
+ * @param[out] outCorrelationDataByteLen Optional. If not NULL, will set to byte length of the mqtt correlation-data property.
  * @return `true` if the correlation data matches the provided correlation ID; otherwise, `false`.
  */
-bool ADU_are_correlation_ids_matching(const mosquitto_property* props, const char* correlationId)
+bool ADU_are_correlation_ids_matching(const mosquitto_property* props, const char* correlationId, char** outCorrelationData, uint16_t* outCorrelationDataByteLen)
```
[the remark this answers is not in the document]
jw-msft: I've made any arguments (in or out) size_t but for locals that need to interact with mosquitto APIs I've kept them as-is.  Let me know if you feel strongly about it and I can revisit.

src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/adu_agent_info_management.c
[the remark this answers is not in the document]
jw-msft: Fixed with commit: [mv corr id test after fn arg validation](https://github.com/Azure/iot-hub-device-update/pull/584/commits/9786d55efcea2b187b79611712c2f6ed92527163)

src/utils/c_utils/src/string_c_utils.c
jw-msft: won't fix. There is unit test for empty src and it's safe due to use of ADUC_Safe_StrCopyN below and it will fail if srcLen was > 0. Anyways, these functions are revamped a bit in a later stack PR for branch that was merged into user/jw-msft/v2-procupd

jw-msft:
merged up to top pr stack that was merged to topic

PR #588, opened 2023-12-06
Add null check on src str in ADUC_AllocStrCopyN

jw-msft:
- Check for NULL for src argument
- Add some inline comments for clarity


src/utils/c_utils/src/string_c_utils.c
```diff
@@ -453,7 +453,7 @@ int ADUC_AllocAndStrCopyN(char** dest, const char* src, size_t srcLen)
     size_t num_copied = 0;
     char* tmpDest = NULL;
 
-    if (dest == NULL || srcLen == 0)
+    if (dest == NULL || src == NULL || srcLen == 0)
```
JeffMill, line 456: return values differ from your ADUC_Safe_StrCopyN which returns number of chars copied.  i can see this potentially causing coding errors.

JeffMill, line 456: start with `*dest = \0'`

JeffMill: (after dest == null check)

src/utils/c_utils/src/string_c_utils.c
```diff
@@ -464,10 +464,13 @@ int ADUC_AllocAndStrCopyN(char** dest, const char* src, size_t srcLen)
         goto done;
     }
 
+    // Uses ADUC_Safe_StrCopyN to prevent buffer overflows and ensure null termination.
```
JeffMill, line 467: not optimal, but would it be worthwhile to be defensive and have this method do a strlen(src) and fail if srcLen > strlen(src)?

src/utils/c_utils/src/string_c_utils.c
```diff
@@ -464,10 +464,13 @@ int ADUC_AllocAndStrCopyN(char** dest, const char* src, size_t srcLen)
...
     num_copied = ADUC_Safe_StrCopyN(
         tmpDest, src, (srcLen + 1) * sizeof(char) /* destByteLen */, srcLen /* numSrcCharsToCopy */);
+
     if (num_copied < srcLen)
     {
+        // it was truncated!
```
JeffMill, line 473: silently truncating a string seems really bad to me.

char path[8 + 1];
// what happens if we access "/this/is" ?!?
ADUC_AllocAndStrCopyN("/this/is/a/full/path", path, sizeof(path))

JeffMill (changes requested):

src/utils/c_utils/src/string_c_utils.c
jw-msft: Good suggestion. I've removed implicit truncation and return an error instead (made return type an int where 0 means success.

src/utils/c_utils/src/string_c_utils.c
jw-msft: done

jw-msft: they both now return int where 0 is success.

src/utils/c_utils/src/string_c_utils.c
jw-msft: This is a great suggestion. Pushed a commit that implements this along with obligatory unit tests. Also, added 4-byte utf-8 test case for duck emoji.

src/extensions/agent_modules/shared/agentinfo_utils/src/adu_agentinfo_utils.c
```diff
@@ -88,19 +88,25 @@ AgentInfoDataFromRetriableOperationContext(ADUC_Retriable_Operation_Context* ret
...
 
-    ADUC_Safe_StrCopyN(
+    if (ADUC_Safe_StrCopyN(
         agentInfoData->ainfoReqMessageContext.correlationId,
         correlationId,
         ARRAY_SIZE(agentInfoData->ainfoReqMessageContext.correlationId),
-        strlen(correlationId));
+        strlen(correlationId)) != 0)
+    {
+        return false;
```
Jerry-GeGarcia, line 106: any tracing valuable here?

src/utils/c_utils/tests/c_utils_ut.cpp
```diff
@@ -384,7 +383,18 @@ TEST_CASE("ADUC_StrNLen")
 
         CHECK(ADUC_StrNLen(testStr.c_str(), max) == max);
     }
+    SECTION("Check duck emoji codepoint")
+    {
+        // The "duck" emoji is unicode codepoint: U+1F986
+        // which falls in the range of U+10000 - U+10FFFF, so in utf-8 it is encoded in the 4-byte form:
+        // 11110xxx 10xxxxxx 10xxxxxx 10xxxxxx
+        CHECK(2 == ADUC_StrNLen("🦆", 2));
```
Jerry-GeGarcia, line 391: >🦆 [](http://example.com/codeflow?start=33&length=2)

:)

Jerry-GeGarcia (dismissed):
:shipit:

src/extensions/agent_modules/shared/agentinfo_utils/src/adu_agentinfo_utils.c
jw-msft: there is some value. added.

Jerry-GeGarcia (approved):
:shipit:

src/utils/c_utils/src/string_c_utils.c
```diff
@@ -417,26 +420,35 @@ bool MallocAndSubstr(char** target, char* source, size_t len)
  *
  * @param dest The destination buffer.
  * @param src The source string to be copied.
- * @param destByteLen The size of the destination buffer in bytes.
+ * @param destByteLen The size of the destination buffer in bytes, including plus 1 for the null terminator.
  * @param numSrcCharsToCopy The number of source chars to copy. It will be truncated and null-terminated if >= destByteLen.
- * @return size_t The number of src chars copied. If < numSrcCharsToCopy, then it was truncated.
+ * @return int zero on success, or a negative number if there was an error.
```
JeffMill, line 425: why'd you choose zero for success? I can easily see this occurring?

```c
if (!ADUC_Safe_StrCopyN(dest, src, len, tocopy)) {
  return false;
}

**Review by JeffMill — CHANGES_REQUESTED** [review]

File: src/utils/c_utils/src/string_c_utils.c (hunk 3 above)
Comment (jw-msft) [reply src/utils/c_utils/src/string_c_utils.c]: caller is supposed to use return value to check for truncation, but I think you're right that just returning an error here would be better.  fixed.

File: src/utils/c_utils/src/string_c_utils.c (hunk 6 above)
Comment (jw-msft) [reply src/utils/c_utils/src/string_c_utils.c]: Changed to return bool true as I think your suggestion is more natural here.

However, due to C's implicit type coercion, returning int 0 is just as fraught with danger as return bool true, depending on if whoever is using it assumes one or the other (and ignores the code and/or doxygen comments surfacing in the IDE or editor).

There's no gcc compiler flag I know of that will warn/error if coercion is going on, but clang-tidy should catch it via https://releases.llvm.org/13.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/readability-implicit-bool-conversion.html

File: src/extensions/agent_modules/shared/agentinfo_utils/src/adu_agentinfo_utils.c [hunk 7]
```diff
@@ -92,20 +92,18 @@ AgentInfoDataFromRetriableOperationContext(ADUC_Retriable_Operation_Context* ret
...
 bool AgentInfoData_SetCorrelationId(ADUC_AgentInfo_Request_Operation_Data* agentInfoData, const char* correlationId)
 {
-    int ret = -1;
     if (agentInfoData == NULL || correlationId == NULL)
     {
         return false;
     }
 
-    ret = ADUC_Safe_StrCopyN(
+    if (!ADUC_Safe_StrCopyN(
```
JeffMill, line 100: would ordering this api to match [StringCchCopyNA](https://learn.microsoft.com/en-us/windows/win32/api/strsafe/nf-strsafe-stringcchcopyna) be worthwhile?  I like that ordering better as it keeps lengths next to the args they reference.

src/utils/c_utils/src/string_c_utils.c
```diff
@@ -422,50 +422,50 @@ bool MallocAndSubstr(char** target, char* source, size_t len)
  * @param src The source string to be copied.
  * @param destByteLen The size of the destination buffer in bytes, including plus 1 for the null terminator.
  * @param numSrcCharsToCopy The number of source chars to copy. It will be truncated and null-terminated if >= destByteLen.
- * @return int zero on success, or a negative number if there was an error.
+ * @return true on success.
  */
-int ADUC_Safe_StrCopyN(char* dest, const char* src, size_t destByteLen, size_t numSrcCharsToCopy)
+bool ADUC_Safe_StrCopyN(char* dest, const char* src, size_t destByteLen, size_t numSrcCharsToCopy)
```
JeffMill, line 427: "chars" vs "bytes"?

src/utils/c_utils/src/string_c_utils.c
```diff
@@ -422,50 +422,50 @@ bool MallocAndSubstr(char** target, char* source, size_t len)
  * @param src The source string to be copied.
  * @param destByteLen The size of the destination buffer in bytes, including plus 1 for the null terminator.
```
JeffMill, line 423: confusing. does one pass in "sizeof(dest)+1" or are you stating that the buffer size should be at least one more than the desired length?

src/utils/c_utils/src/string_c_utils.c
```diff
@@ -422,50 +422,50 @@ bool MallocAndSubstr(char** target, char* source, size_t len)
...
     }
     *dest = '\0';
 
     if (numSrcCharsToCopy >= destByteLen)
     {
-        return -2;
+        return false;
     }
 
     srcLen = ADUC_StrNLen(src, numSrcCharsToCopy);
```
JeffMill, line 442: nit (minor optimization): could just start copying, and fail if we get to srcLen characters.

src/utils/c_utils/src/string_c_utils.c
```diff
@@ -422,50 +422,50 @@ bool MallocAndSubstr(char** target, char* source, size_t len)
...
 /**
 * @brief Allocate a new str that is a copy of the src string with given max number of char.
 * @param dest The addr of ptr that will point to newly allocated string copy.
 * @param src The source string.
 * @param srcLen The char length of source string (not including null terminator)
-* @return 0 on success, -2 on invalid argument, and -1 for other errors.
+* @return true on success
 */
-int ADUC_AllocAndStrCopyN(char** dest, const char* src, size_t srcLen)
+bool ADUC_AllocAndStrCopyN(char** dest, const char* src, size_t srcLen)
```
JeffMill, line 461: "len" vs. "bytes" vs. "chars" (above)

src/utils/c_utils/src/string_c_utils.c
```diff
@@ -422,50 +422,50 @@ bool MallocAndSubstr(char** target, char* source, size_t len)
...
 {
-    int res = -1;
+    bool res = false;
     char* tmpDest = NULL;
 
     if (dest == NULL || src == NULL || srcLen == 0)
```
JeffMill, line 466: so this isn't allowed?

`AllocAndStrCopyN(&dest, "jeff", 0); // should return allocated `""`?`

JeffMill (approved):
approved with suggestions

src/extensions/agent_modules/shared/agentinfo_utils/src/adu_agentinfo_utils.c
jw-msft: That makes sense to put the relevant counts adjacent. Will send that as a separate PR.

src/utils/c_utils/src/string_c_utils.c
jw-msft: That is confusing in retrospect. Was trying to convey that destByteLen should include capacity for the additional nul-term. Will fix in next PR.

src/utils/c_utils/src/string_c_utils.c
jw-msft: should be named bytes consistently since it could be a char array of utf-8 (or other encoding) encoded chars that each can be up to 4 bytes nowadays (I remember when it was a max of 3).  Will address in follow-up PR.

src/utils/c_utils/src/string_c_utils.c
jw-msft: not currently as this function doesn't allocate on caller's behalf. For that, one can call other existing string_c_utils function above: MallocAndSubstr(&dest, "jeff", 0);

src/extensions/agent_modules/shared/agentinfo_utils/src/adu_agentinfo_utils.c
jw-msft: Opened this to track: [Bug 48174247](https://microsoft.visualstudio.com/OS/_workitems/edit/48174247): [Gen2]Align ADUC_Safe_StrCopyN arg order with StringCchCopyNA

jw-msft: @JeffMill PR to address is here: https://github.com/Azure/iot-hub-device-update/pull/590

src/utils/c_utils/src/string_c_utils.c
jw-msft: Should be resolved with this PR: https://github.com/Azure/iot-hub-device-update/pull/590

src/utils/c_utils/src/string_c_utils.c
jw-msft: fixed here: https://github.com/Azure/iot-hub-device-update/pull/590

PR #590, opened 2023-12-14
Align ADUC_Safe_StrCopyN arg order to StringCchCopyNA

jw-msft:
- [align ADUC_Safe_StrCopyN arg order to StringCchCopyNA](https://github.com/Azure/iot-hub-device-update/commit/0325d011c553598a254ea4dbf606999c42c44935)
- [clang-format](https://github.com/Azure/iot-hub-device-update/commit/883cd019a08c9067783acc13ac6dc5725ea226e6)
- Make arg naming consistent: dest, destByteLen, src, srcByteLen

PR #591, opened 2023-12-18
optimize ADUC_Safe_StrCopyN

jw-msft:
( stacked on PR https://github.com/Azure/iot-hub-device-update/pull/590 )
- copy to dest directly and check for EOS before srcIndex exceeds srcByteLen instead of StrLenN + memcpy
- update unit test case

### DIFF
--- a/src/utils/c_utils/src/string_c_utils.c
+++ b/src/utils/c_utils/src/string_c_utils.c
@@ -384,8 +384,8 @@ bool IsNullOrEmpty(const char* str)
  * @brief Allocates memory for @p target and copy @p len characters of @p source into @p target buffer.
  *
  * @param[out] target Output string
- * @param source Source string
  * @param len Length of string to copy.
+ * @param source Source string
  * @return bool Returns true is success.
  */
 bool MallocAndSubstr(char** target, char* source, size_t len)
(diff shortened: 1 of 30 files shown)


Nox-MSFT (approved):

src/utils/c_utils/src/string_c_utils.c
```diff
@@ -440,13 +442,17 @@ bool ADUC_Safe_StrCopyN(char* dest, size_t destByteLen, const char* src, size_t
...
+    for (srcIndex = 0; (srcIndex < srcByteLen) && (src[srcIndex] != '\0'); ++srcIndex)
     {
-        return false;
+        *dest++ = src[srcIndex];
     }
+    *dest = '\0';
 
-    memcpy(dest, src, srcByteLen);
-    dest[srcByteLen] = '\0';
+    if ((srcIndex < srcByteLen) && (src[srcIndex] == '\0'))
```
JeffMill, line 451: see my comment in another PR you have out.

JeffMill (changes requested):

src/utils/c_utils/src/string_c_utils.c
jw-msft: I'm inclined to "won't fix" this due to the following:
- The purpose is to allow caller to specify a substring or num bytes of a non-null-terminated source string.
- Caller knows the character encoding so can translate to number of bytes for the encoded characters.
- Also, can't assume the src string is null-terminated but this'll make sure the dest buffer is.
- Here is the UT that exercises this branch:
https://github.com/Azure/iot-hub-device-update/blob/ca5643f81c66905429c110b9d8da7e81f491feb4/src/utils/c_utils/tests/c_utils_ut.cpp#L507

jw-msft:
Closing in favor of PR on the top of stack: https://github.com/Azure/iot-hub-device-update/pull/593

PR #592, opened 2023-12-18
Add ADUC_CopyIf and use for safe StrCopyN impl

jw-msft:
( stacked on PR https://github.com/Azure/iot-hub-device-update/pull/591 )
- [add CopyIf to string_c_utils](https://github.com/Azure/iot-hub-device-update/commit/8e5d810893e9a61b2ff484b8ce192c90495ee134)
- [use ADUC_CopyIf for ADUC_Safe_StrCopyN impl](https://github.com/Azure/iot-hub-device-update/commit/2b73ffe7c32f3fb3a3906033789fd74563cad4eb)

PR #593, opened 2023-12-18
Add workaround for du service not preserving exact correlation id string

jw-msft:
(stacked on PR https://github.com/Azure/iot-hub-device-update/pull/592 )
- In correlation id match, use ADUC_CopyIf to cpy non dashes as service sends correlation id with dashes despite request having no dashes


src/extensions/agent_modules/shared/mosquitto_utils/src/adu_mosquitto_utils.c
```diff
@@ -140,7 +146,25 @@ bool ADU_are_correlation_ids_matching(
 
     if (outCorrelationData != NULL)
     {
-        if (ADUC_AllocAndStrCopyN(outCorrelationData, responseCorrelationId, (size_t)responseCorrelationIdLen) != 0)
+        // TODO: Remove this workaround once adu service deploys fix that ensures exact correlation id string agent sent is
```
Jerry-GeGarcia, line 149: is there a bug tracking the service side fix?

Nox-MSFT (approved):

Nox-MSFT (approved):

Nox-MSFT (approved):

src/extensions/agent_modules/shared/mosquitto_utils/src/adu_mosquitto_utils.c
```diff
@@ -24,6 +24,11 @@
 // keep this last to avoid interfering with system headers
 #include <aduc/aduc_banned.h>
 
+static bool not_a_dash(char c)
```
JeffMill, line 27: seems oddly specific (and not really reusable) to create a "not a dash" method.  just add an inline lambda at the source.

src/extensions/agent_modules/shared/mosquitto_utils/src/adu_mosquitto_utils.c
JeffMill: please add the # to the comment.

src/extensions/agent_modules/shared/mosquitto_utils/src/adu_mosquitto_utils.c
```diff
@@ -140,7 +146,25 @@ bool ADU_are_correlation_ids_matching(
 
     if (outCorrelationData != NULL)
     {
-        if (ADUC_AllocAndStrCopyN(outCorrelationData, responseCorrelationId, (size_t)responseCorrelationIdLen) != 0)
+        // TODO: Remove this workaround once adu service deploys fix that ensures exact correlation id string agent sent is
+        // in the response message. Currently, it parses the UUID without dashes and then introduces dashes in the response
+        // message.
+        char correlation_id_no_dash[37] = "";
```
JeffMill, line 152: don't init out params.

presumably 37 is GUID_LEN + 1?  If so, use a define instead of a magic number.  If it's coincidental that it's GUID_LEN, then create a "CORRELATION_ID_LEN" define.

src/extensions/agent_modules/shared/mosquitto_utils/src/adu_mosquitto_utils.c
```diff
@@ -140,7 +146,25 @@ bool ADU_are_correlation_ids_matching(
...
+        char* destEnd = NULL;
+        if (!ADUC_CopyIf(
```
JeffMill, line 154: presumably this ADU's "C" version of std::copy_if ?

src/extensions/agent_modules/shared/mosquitto_utils/src/adu_mosquitto_utils.c
JeffMill, line 149: reviewing now -- line 140 seems really dangerous and potential for a 0-day.  if read_binary returns binary data, how do we know:

1. that it's a string
2. that it's null-terminated?

src/extensions/agent_modules/shared/mosquitto_utils/src/adu_mosquitto_utils.c
```diff
@@ -140,7 +146,25 @@ bool ADU_are_correlation_ids_matching(
...
+            correlation_id_no_dash /* dest */,
+            sizeof(correlation_id_no_dash) /* destByteLen */,
+            responseCorrelationId /* src */,
+            (size_t)responseCorrelationIdLen /* srcByteLen */,
+            not_a_dash /* copyPred */,
```
JeffMill, line 159: as mentioned before, use an inline lambda fn.

src/extensions/agent_modules/shared/mosquitto_utils/src/adu_mosquitto_utils.c
```diff
@@ -140,7 +146,25 @@ bool ADU_are_correlation_ids_matching(
...
+            &destEnd /* outDestEnd */))
+        {
+            goto done;
+        }
+
+        byteLenNoDashes = (size_t)(destEnd - correlation_id_no_dash);
+
+        if (!ADUC_AllocAndStrCopyN(outCorrelationData, correlation_id_no_dash, byteLenNoDashes))
```
JeffMill, line 167: this is horribly confusing to me, and likely going to cause some security issue at some point if not improved.

I see 

* byteLenNoDashes
* outCorrelationDataByteLen
* responseCorrelationIdLen
* n
* sizeof(correlation_id_no_dash)

how is one supposed to track what all of these indicate, and which indicate a buffer size vs a buffersize (minus or including) a null terminator or which match which buffer.

We should look into a better way to do this in C even if means we use the STRING_HANDLE and deal with extra allocations.

JeffMill (changes requested):

src/extensions/agent_modules/shared/mosquitto_utils/src/adu_mosquitto_utils.c
JeffMill: oh crap, it's C code.

src/extensions/agent_modules/shared/mosquitto_utils/src/adu_mosquitto_utils.c
JeffMill: meh, c code.

src/extensions/agent_modules/shared/mosquitto_utils/src/adu_mosquitto_utils.c
jw-msft: Here it is @Jerry-GeGarcia  and @JeffMill : [Bug 48216724](https://microsoft.visualstudio.com/OS/_workitems/edit/48216724): [gen2] service returns modified correlation id in the response

src/extensions/agent_modules/shared/mosquitto_utils/src/adu_mosquitto_utils.c
jw-msft: yeah, I'd definitely use a lambda if c++

src/extensions/agent_modules/shared/mosquitto_utils/src/adu_mosquitto_utils.c
jw-msft: Yes, basically.

src/extensions/agent_modules/shared/mosquitto_utils/src/adu_mosquitto_utils.c
jw-msft: Yes, good catch!  

To properly validate, it needs to be verifying that the bytes patterns are those of valid utf-8 encodings. Also, after such verification, it cannot assume the utf-8 is null-terminated. However, we /are/ given the byte length (`responseCorrelationIdLen`) so the latter could be avoided.

I'm going to side-step the copying and output parameters because it's only needed for tracing to logs. Instead, I will replace with a trace (and add comments not to trace the binary data that could possibly NOT be utf-8) with:

a static trace that just says the trusted `expectededCorrelationId` string did not match the untrusted binary data correlation-data.

src/extensions/agent_modules/shared/mosquitto_utils/src/adu_mosquitto_utils.c
jw-msft: yeah, I'm removing these out parameters now for security and other reasons.

PR #595, opened 2023-12-20
fix idlewait interval, jitter and reconnect wait

jw-msft:
- [fix idlewait interval, jitter and reconnect wait](https://github.com/Azure/iot-hub-device-update/commit/4d94bd3a3a8dde3d5bf0864f6de1e937334bd3e7)

### DIFF
--- a/src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/upd_request_operation.c
+++ b/src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/upd_request_operation.c
@@ -11,7 +11,8 @@
 #include <du_agent_sdk/agent_module_interface.h> // ADUC_AGENT_MODULE_HANDLE
 
 // Forward declarations
-void Adu_ProcessUpdate(ADUC_Update_Request_Operation_Data* updateData, ADUC_Retriable_Operation_Context* retriableOperationContext);
+void Adu_ProcessUpdate(
+    ADUC_Update_Request_Operation_Data* updateData, ADUC_Retriable_Operation_Context* retriableOperationContext);
 time_t ADUC_Retry_Delay_Calculator(
     int additionalDelaySecs,
     unsigned int retries,
@@ -44,7 +45,7 @@ bool UpdateRequestOperation_CancelOperation(ADUC_Retriable_Operation_Context* co
 
     context->lastFailureTime = now;
     context->nextExecutionTime = now + 180; // TODO: use retryParams.
-    AduUpdUtils_TransitionState(ADU_UPD_STATE_IDLEWAIT, updData);
+    AduUpdUtils_TransitionState(ADU_UPD_STATE_IDLEWAIT, updData, context);
     return true;
 }
 
@@ -69,7 +70,7 @@ bool UpdateRequestOperation_Complete(ADUC_Retriable_Operation_Context* context)
 
     context->nextExecutionTime = ADUC_GetTimeSinceEpochInSeconds() + 30;
 
-    AduUpdUtils_TransitionState(ADU_UPD_STATE_IDLEWAIT, updData);
+    AduUpdUtils_TransitionState(ADU_UPD_STATE_IDLEWAIT, updData, context);
     return true;
 }
 
@@ -115,7 +116,7 @@ bool UpdateRequestOperation_DoRetry(ADUC_Retriable_Operation_Context* context, c
         Log_Debug("Max retry of %d exceeded. Failing.", retryCfg->maxRetries);
 
         context->lastFailureTime = now;
-        AduUpdUtils_TransitionState(ADU_UPD_STATE_IDLEWAIT, data);
+        AduUpdUtils_TransitionState(ADU_UPD_STATE_IDLEWAIT, data, context);
     }
     else
     {
@@ -131,7 +132,7 @@ bool UpdateRequestOperation_DoRetry(ADUC_Retriable_Operation_Context* context, c
 
         context->nextExecutionTime = newTime;
 
-        AduUpdUtils_TransitionState(ADU_UPD_STATE_RETRYWAIT, data);
+        AduUpdUtils_TransitionState(ADU_UPD_STATE_RETRYWAIT, data, context);
     }
 
     return true;
@@ -302,7 +303,7 @@ static bool SendUpdateRequest(ADUC_Retriable_Operation_Context* context, time_t 
         goto done;
     }
 
-    AduUpdUtils_TransitionState(ADU_UPD_STATE_REQUESTING, updData);
+    AduUpdUtils_TransitionState(ADU_UPD_STATE_REQUESTING, updData, context);
     ADUC_Retriable_Set_State(context, ADUC_Retriable_Operation_State_InProgress);
 
     context->lastExecutionTime = nowTime;
@@ -324,7 +325,8 @@ done:
     return opSucceeded;
 }
 
-static bool SendUpdateResultsRequest(ADUC_Retriable_Operation_Context* context, ADUC_Update_Request_Operation_Data* updData, time_t nowTime)
+static bool SendUpdateResultsRequest(
+    ADUC_Retriable_Operation_Context* context, ADUC_Update_Request_Operation_Data* updData, time_t nowTime)
 {
     bool opSucceeded = false;
     int mqtt_res = 0;
@@ -426,7 +428,7 @@ static bool SendUpdateResultsRequest(ADUC_Retriable_Operation_Context* context, 
         goto done;
     }
 
-    AduUpdUtils_TransitionState(ADU_UPD_STATE_REQUESTING, updData);
+    AduUpdUtils_TransitionState(ADU_UPD_STATE_REQUESTING, updData, context);
     ADUC_Retriable_Set_State(context, ADUC_Retriable_Operation_State_InProgress);
 
     context->lastExecutionTime = nowTime;
@@ -448,7 +450,8 @@ done:
     return opSucceeded;
 }
 
-static void HandleReportResults(ADUC_Retriable_Operation_Context* context, ADUC_Update_Request_Operation_Data* updData, time_t nowTime)
+static void HandleReportResults(
+    ADUC_Retriable_Operation_Context* context, ADUC_Update_Request_Operation_Data* updData, time_t nowTime)
 {
     if (nowTime > context->nextExecutionTime)
     {
@@ -461,7 +464,7 @@ static void HandleReportResults(ADUC_Retriable_Operation_Context* context, ADUC_
             context->attemptCount = 0;
             context->lastFailureTime = nowTime;
             context->nextExecutionTime = nowTime + 60; // TODO: add random jitter and use retry configuration.
-            AduUpdUtils_TransitionState(ADU_UPD_STATE_IDLEWAIT, updData);
+            AduUpdUtils_TransitionState(ADU_UPD_STATE_IDLEWAIT, updData, context);
         }
         else
         {
@@ -481,13 +484,14 @@ static void HandleReportResults(ADUC_Retriable_Operation_Context* context, ADUC_
                 free(updData->reportingJson);
                 updData->reportingJson = NULL;
                 ADUC_StateStore_SetReportResultsAck(false);
-                AduUpdUtils_TransitionState(ADU_UPD_STATE_REPORT_RESULTS_ACK, updData);
+                AduUpdUtils_TransitionState(ADU_UPD_STATE_REPORT_RESULTS_ACK, updData, context);
             }
         }
     }
 }
 
-static void HandleReportResultsAck(ADUC_Retriable_Operation_Context* context, ADUC_Update_Request_Operation_Data* updData, time_t nowTime)
+static void HandleReportResultsAck(
+    ADUC_Retriable_Operation_Context* context, ADUC_Update_Request_Operation_Data* updData, time_t nowTime)
 {
     if (nowTime > context->nextExecutionTime)
     {
@@ -499,7 +503,7 @@ static void HandleReportResultsAck(ADUC_Retriable_Operation_Context* context, AD
             context->attemptCount = 0;
             context->lastFailureTime = nowTime;
             context->nextExecutionTime = nowTime + 60; // TODO: add random jitter and use retry configuration.
-            AduUpdUtils_TransitionState(ADU_UPD_STATE_IDLEWAIT, updData);
+            AduUpdUtils_TransitionState(ADU_UPD_STATE_IDLEWAIT, updData, context);
         }
         else
         {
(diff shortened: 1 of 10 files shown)


src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/upd_request_operation.c
```diff
@@ -461,7 +464,7 @@ static void HandleReportResults(ADUC_Retriable_Operation_Context* context, ADUC_
             context->attemptCount = 0;
             context->lastFailureTime = nowTime;
             context->nextExecutionTime = nowTime + 60; // TODO: add random jitter and use retry configuration.
-            AduUpdUtils_TransitionState(ADU_UPD_STATE_IDLEWAIT, updData);
+            AduUpdUtils_TransitionState(ADU_UPD_STATE_IDLEWAIT, updData, context);
```
nihemstr, line 467: nit: add the + 60 to a configuration that we use for the nextExecution time. 

Also commented out line on line 459. Do we need it still?

src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/upd_request_operation.c
```diff
@@ -481,13 +484,14 @@ static void HandleReportResults(ADUC_Retriable_Operation_Context* context, ADUC_
                 free(updData->reportingJson);
                 updData->reportingJson = NULL;
                 ADUC_StateStore_SetReportResultsAck(false);
-                AduUpdUtils_TransitionState(ADU_UPD_STATE_REPORT_RESULTS_ACK, updData);
+                AduUpdUtils_TransitionState(ADU_UPD_STATE_REPORT_RESULTS_ACK, updData, context);
```
nihemstr, line 487: in line freeing of updData reportingJson is being freed inline here. Do we need to free it if the SendUpdateResultsRequest fails? 

Also using context->retryParams here for the maxDelaySecs but max_retries doesn't use retry params. See line 459: 

> const unsigned int max_retries = context->retryParams[ADUC_RETRY_PARAMS_INDEX_DEFAULT].maxRetries

src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/upd_request_operation.c
```diff
@@ -481,13 +484,14 @@ static void HandleReportResults(ADUC_Retriable_Operation_Context* context, ADUC_
...
             }
         }
     }
 }
 
-static void HandleReportResultsAck(ADUC_Retriable_Operation_Context* context, ADUC_Update_Request_Operation_Data* updData, time_t nowTime)
+static void HandleReportResultsAck(
+    ADUC_Retriable_Operation_Context* context, ADUC_Update_Request_Operation_Data* updData, time_t nowTime)
```
nihemstr, line 494: input validation on updData just in case unless expecting to fail loudly (Please just resolve these if we're expecting a loud failure in the void functions. Just trying to cover the bases)

src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/upd_request_operation.c
```diff
@@ -499,7 +503,7 @@ static void HandleReportResultsAck(ADUC_Retriable_Operation_Context* context, AD
             context->attemptCount = 0;
             context->lastFailureTime = nowTime;
             context->nextExecutionTime = nowTime + 60; // TODO: add random jitter and use retry configuration.
-            AduUpdUtils_TransitionState(ADU_UPD_STATE_IDLEWAIT, updData);
+            AduUpdUtils_TransitionState(ADU_UPD_STATE_IDLEWAIT, updData, context);
```
nihemstr, line 506: #define for the + 60. Same question about the nowTime verses using the time we transition state and/or send the report to prevent drift from caller execution to this function.

src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/upd_request_operation.c
```diff
@@ -511,7 +515,7 @@ static void HandleReportResultsAck(ADUC_Retriable_Operation_Context* context, AD
                 context->lastSuccessTime = nowTime;
                 context->nextExecutionTime = nowTime + 60; // TODO: add random jitter and use retry configuration.
                 ADUC_StateStore_SetReportResultsAck(false);
-                AduUpdUtils_TransitionState(ADU_UPD_STATE_IDLEWAIT, updData);
+                AduUpdUtils_TransitionState(ADU_UPD_STATE_IDLEWAIT, updData, context);
```
nihemstr, line 518: Do we have a bug or anything else to track the todos for adding random jitter and using the retry configuration?

src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/upd_request_operation.c
```diff
@@ -572,7 +576,7 @@ bool UpdateRequestOperation_DoWork(ADUC_Retriable_Operation_Context* context)
         {
             Log_Debug("exiting IDLEWAIT state: %lu > %lu", nowTime, context->nextExecutionTime);
 
-            AduUpdUtils_TransitionState(ADU_UPD_STATE_READY, updData);
+            AduUpdUtils_TransitionState(ADU_UPD_STATE_READY, updData, context);
```
nihemstr, line 579: No check on the updData after call to UpdateData_FromOperationContext(). We should standardize the checks on these. Either we manually check if context is not null and context->data is not null or we check that context is not null and then after the call to UpdateData_FromOperationContext() we check the value of updData (this seems like the most reasonable) for all functions.

src/extensions/agent_modules/shared/upd_utils/src/adu_upd_utils.c
```diff
@@ -127,4 +128,18 @@ void AduUpdUtils_TransitionState(ADU_UPD_STATE newState, ADUC_Update_Request_Ope
         newState,
         AduUpdState_str(newState));
     updateData->updState = newState;
+
+    if (newState == ADU_UPD_STATE_IDLEWAIT)
```
nihemstr, line 132: Same no check on inputs before utilizing the values in the Log_Info struct

src/extensions/agent_modules/shared/upd_utils/src/adu_upd_utils.c
```diff
@@ -127,4 +128,18 @@ void AduUpdUtils_TransitionState(ADU_UPD_STATE newState, ADUC_Update_Request_Ope
...
+    {
+        srand(time(NULL));
+        int jitter_seconds = (rand() % 121) + 60; // between 1 and 3 minutes of jitter
```
nihemstr, line 135: Should be a #define centralized in a seperate .h file for all these random numbers.

src/extensions/agent_modules/shared/upd_utils/src/adu_upd_utils.c
```diff
@@ -127,4 +128,18 @@ void AduUpdUtils_TransitionState(ADU_UPD_STATE newState, ADUC_Update_Request_Ope
...
+        time_t now = ADUC_GetTimeSinceEpochInSeconds();
+        context->nextExecutionTime = now + context->pollingIntervalInSeconds + jitter_seconds;
```
nihemstr, line 137: We need toc heck if context is null before assigning/accessing variables unless we want the boom to do the error checking for us.

src/utils/parson_json_utils/src/parson_json_utils.c
```diff
@@ -236,7 +236,7 @@ bool ADUC_JSON_GetUnsignedIntegerField(const JSON_Value* jsonValue, const char*
     val = json_object_dotget_number(jsonObj, jsonFieldName);
 
     // Note: check value is not negative and is a whole number for safe casting to an unsigned int
-    if (val < 0 || ((int)val) != val)
+    if (val <= 0 || ((int)val) != val)
```
nihemstr, line 239: Zed is a valid non negative whole number. This would prevent any call to get a value of 0 from a json structure.

src/utils/retry_utils/inc/aduc/retry_utils.h
```diff
@@ -55,6 +55,9 @@ EXTERN_C_BEGIN
 #define DEFAULT_ENR_REQ_OP_MAX_JITTER_PERCENT (60)
 #define SETTING_KEY_ENR_REQ_OP_MAX_JITTER_PERCENT "maxJitterPercent"
 
+#define SETTING_KEY_POLLING_INTERVAL_SECONDS "pollingIntervalInSeconds"
+#define ADUC_DEFAULT_POLLING_INTERVAL_SECONDS (5 * 60)
```
nihemstr, line 59: This might not be the right PR but aren't these apart of the configuration file (like the one in the example?). If so these access functions and the defaults should be inside of config_utils not the retry_utils

src/utils/retry_utils/src/retry_utils.c
```diff
@@ -186,14 +186,15 @@ int RetryUtils_GetRetryParamsMapSize()
...
-    ADUC_Retriable_Operation_Context* context, JSON_Value* agentJsonValue, int retryParamsMapSize)
+    ADUC_Retriable_Operation_Context* context, JSON_Value* agentRetrySettingsJsonValue, int retryParamsMapSize)
 {
     const char* infoFormatString = "Failed to read '%s.%s' from agent config. Using default value (%d)";
 
     for (int i = 0; i < retryParamsMapSize; i++)
     {
         ADUC_Retry_Params* params = &context->retryParams[i];
-        JSON_Value* retryParams = json_object_dotget_value(json_object(agentJsonValue), s_retryParamsMap[i].name);
+        JSON_Value* retryParams =
```
nihemstr, line 196: immediately casting agentRetrySettingsJsonValue to json_object without checking inputs nor the object it's casted to.

nihemstr, line 196: int i should be size_t i and the input retryParamsMapSize should be size_t to standardize types across the code base and simplify the windows integration.

nihemstr, line 196: Might be overloading the function but should be using dothas_ to determine if we need to add the default configuration.

If the configuration has the value but it is invalid, we need to warn the customer instead of assuming it just doesn't exist.

nihemstr, line 196: This function is used to interact with the agent_info->agentJsonValue which I think is our configuration file if I traced it right. 

We should move the functions for parsing in the configuration into config_utils and add the defaults to config_utils so we can centralize all of the configuration work + unit tests for reading and setting the value.

src/utils/retry_utils/src/retry_utils.c
```diff
@@ -259,5 +260,18 @@ void ReadRetryParamsArrayFromAgentConfigJson(
             value = DEFAULT_ENR_REQ_OP_MAX_JITTER_PERCENT;
         }
         params->maxJitterPercent = (double)value;
+
+        //
+        // Get the update polling interval retry setting, or default.
+        //
+        if (!ADUC_JSON_GetUnsignedIntegerField(
```
nihemstr, line 267: The change to ADUC_JSON_GetUnsignedIntegerField() does not allow zero but if I am a customer and I want to eliminate retires it would make sense for me to configure the field with a zed value. 

Same for all of the ones below.

src/utils/retry_utils/src/retry_utils.c
```diff
@@ -259,5 +260,18 @@ void ReadRetryParamsArrayFromAgentConfigJson(
...
+                agentRetrySettingsJsonValue, SETTING_KEY_POLLING_INTERVAL_SECONDS, &context->pollingIntervalInSeconds))
```
nihemstr, line 268: There might be a case where polling should never be zero to prevent scheduler domination so would make sense to have a separate check for that and logging that information (eg Log_Info("Polling interval in configuration is zero. Minimum polling interval is default value of: %lu", ADUC_DEFAULT_POLLING_INTERVAL_SECONDS) 

Or you could make a definition for the "safe" minimum polling interval (maybe call it 10% of our default ?)  so customers can still make it faster / slower but they're not going to bork everything by making it zero or like 1 or something. 

If you follow Nyquist's sampling law the system should be polling at 3x the rate of the system. So if the lowest delay we have is 60 seconds a recommended sampling/ scheduling window would be 20 s. Just throw math at the problem and it might go away.

src/utils/retry_utils/src/retry_utils.c
```diff
@@ -259,5 +260,18 @@ void ReadRetryParamsArrayFromAgentConfigJson(
...
+        {
+            Log_Warn(
+                infoFormatString, "", SETTING_KEY_POLLING_INTERVAL_SECONDS, ADUC_DEFAULT_POLLING_INTERVAL_SECONDS);
+
+            context->pollingIntervalInSeconds = ADUC_DEFAULT_POLLING_INTERVAL_SECONDS;
+        }
+        Log_Info("Read Update Polling Interval of %u from config", context->pollingIntervalInSeconds);
```
nihemstr, line 275: nit: It makes it easier to search through the code base for traces if these are inline instead of using a const char* multiple times in different log statements.

nihemstr (approved):
Approved with comments on input validation (pending understanding), configuration function locations, and magic numbers.

src/utils/parson_json_utils/src/parson_json_utils.c
jw-msft: actually, parson returns 0 on failure for the various "get number" APIs, which is why it was not setting to default setting when it was missing in the du-config.json.

That said, I will switch to using json_object_has_value() on these, which is better than using 0

src/extensions/agent_modules/shared/upd_utils/src/adu_upd_utils.c
jw-msft: I would actually prefer it crashes so it's discoverable and can get a core dump as it should never be NULL here.

src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/upd_request_operation.c
jw-msft: that would clutter the code with checks. If initialization of the context objects was done correctly, and the user obj passed to mosquitto_new was good, then these cannot be NULL unless there is something very wrong, in which case we would want to know about that right away by having it crash

src/extensions/agent_modules/cloud_communication/eg_mqtt_broker_client/src/upd_request_operation.c
jw-msft: This is tracking fixing all of these holistically (things outside the scope of this PR): [Bug 48124671](https://microsoft.visualstudio.com/OS/_workitems/edit/48124671): [gen2] Simplify timeout/retry config settings and use them instead of hard-coded defaults.

PR #596, opened 2023-12-21
Added exit handler to ensure no double free in workflow_get_workfolder when parent workfolder is used

nihemstr:
Repro information and log output in bug here: https://microsoft.visualstudio.com/OS/_workitems/edit/48256773

### DIFF
--- a/src/utils/workflow_utils/src/workflow_utils.c
+++ b/src/utils/workflow_utils/src/workflow_utils.c
@@ -1887,8 +1887,15 @@ char* workflow_get_root_sandbox_dir(const ADUC_WorkflowHandle handle)
 
 done:
 
-    free(tempRet);
-    free(pwf);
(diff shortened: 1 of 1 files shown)


Jerry-GeGarcia (dismissed):

src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -1868,6 +1868,7 @@ char* workflow_get_root_sandbox_dir(const ADUC_WorkflowHandle handle)
         }
 
         tempRet = pwf;
+        pwf = NULL;
         Log_Debug("Using parent workfolder: '%s'", pwf);
```
jw-msft, line 1872: Log should happen before transfer ownership, or use tempRet instead of pwf here.

jw-msft (approved):

PR #601, opened 2024-01-18
use workflowId top-level prop in upd_resp

jw-msft:
- Change upd_resp payload parsing to use "workflowId" at top-level instead of workflow.id
- Remove workflow.action that no longer exists
- Update workflow util UTs


src/utils/workflow_utils/src/workflow_utils.c
```diff
@@ -2508,7 +2508,7 @@ const char* workflow_peek_id(ADUC_WorkflowHandle handle)
     if (id == NULL)
     {
         // Return 'workflow.id' from Action json data.
```
Nox-MSFT, line 2510: Nit: update comment here and in function comment also.

Nox-MSFT (approved):
Approved with comment.

Issue #609, opened 2024-03-07
Ignore duplicate deployment workflow not working as expected


JoergZeidler:
During long-term testing I encountered an issue with the deployment status.

## Expected Behavior
Deployment status should not change even after a token expiry refresh.

## Current Behavior
Deployment status changed from Successfully back to In Progress.

## Steps to Reproduce

1. Prepare device with software version "A"
2. Start an update via IotHub to software version "B"
3. software version "B" successfully deployed on device 
3.1. current update state is succeeded 
3.2 deployment status is succeeded (module twin: reported property  "state": 0)
4. wait at least 1h
5. check deployment status: 
5.1. deployments status changed to "**in progress"** (module twin: reported property  "state": 6)   **-> NOK**
5.1. current update state still succeeded 

### Device Information
DU Agent Version 1.1.0

## Logs
token expiry connection refresh changed deployment status to "in progress"

[log.txt](https://github.com/Azure/iot-hub-device-update/files/14524430/log.txt)

cutout:

**Ignoring duplicate deployment** 6409ddb9d8ce39ade510df4a-20242685040, action 3 
```
[ADUC_Workflow_HandleUpdateAction:714]
PropertyUpdated event handler completed. [ADUC_Workflow_HandlePropertyUpdate:638]
Queueing message (t:1, c:0xff3f1c40, m:{"deviceUpdate":{"__t":"c","service":{"value":{"workflow":{"action":3,"id":"6409ddb9d8ce39ade>
OrchestratorPropertyUpdateCallback ended [OrchestratorUpdateCallback:517]
Sending D2C message (t:0, retries:0). [ProcessMessage:390]
Sending D2C message:
 {"deviceUpdate":{"__t":"c","agent":{"lastInstallResult":{"stepResults":null,"resultCode":1,"extendedResultCodes":"00000000","resultDetails":null},**"state":6,"**workflow":{"acti>
 2024-03-06T09:41:32.1685Z 1276[1276] [D] Sending D2C message (t:1, retries:0). [ProcessMessage:390]
```

## Additional Information
PR to fix the issue

JoergZeidler:
I found another use case where the deployment status will be switched back to "in Progress".
Use Case "consciously forcing a failed state"
- agent reports in this case correct state "255"
- after 1h
- token expired: ->  agent reports state "6" -> NOK
```
[...]

PR #610, opened 2024-03-07
fix ignore duplicate deployment use case by set workflow to idle state

JoergZeidler:
see Issue: https://github.com/Azure/iot-hub-device-update/issues/609

### DIFF
--- a/src/adu_workflow/src/agent_workflow.c
+++ b/src/adu_workflow/src/agent_workflow.c
@@ -672,7 +672,10 @@ void ADUC_Workflow_HandleUpdateAction(ADUC_WorkflowData* workflowData)
     if (workflow_isequal_id(workflowData->WorkflowHandle, workflowData->LastCompletedWorkflowId)
         && !workflow_get_force_update(workflowData->WorkflowHandle))
     {
+        char* updateId = workflow_get_expected_update_id_string(workflowData->WorkflowHandle);
         Log_Debug("Ignoring duplicate deployment %s, action %d", workflowData->LastCompletedWorkflowId, desiredAction);
+        ADUC_Workflow_SetInstalledUpdateIdAndGoToIdle(workflowData, updateId);
+        free(updateId);
         goto done;
     }
 


src/adu_workflow/src/agent_workflow.c
```diff
@@ -672,7 +672,10 @@ void ADUC_Workflow_HandleUpdateAction(ADUC_WorkflowData* workflowData)
     if (workflow_isequal_id(workflowData->WorkflowHandle, workflowData->LastCompletedWorkflowId)
         && !workflow_get_force_update(workflowData->WorkflowHandle))
     {
+        char* updateId = workflow_get_expected_update_id_string(workflowData->WorkflowHandle);
         Log_Debug("Ignoring duplicate deployment %s, action %d", workflowData->LastCompletedWorkflowId, desiredAction);
+        ADUC_Workflow_SetInstalledUpdateIdAndGoToIdle(workflowData, updateId);
```
jw-msft, line 677: This will cause it to report the results approximately every 40 minutes, or whatever the token expiration is. This will cause it to quickly consume message quota--although it would help it to go from in-progress state to a terminal state quicker, it does not address the root issue and has negative side-effect of unnecessarily eating message quota.

The root issue is fixed by PR #659 (just waiting on @nihemstr to address the blocking comment I have), which is to avoid setting to in-progress when processing the root key and to set it to in-progress only once it starts processing an update, which was the previous behavior pre-rootkey package.

If possible, please test out PR #659 to see if addresses the issue in issue https://github.com/Azure/iot-hub-device-update/issues/609

JoergZeidler: ok, thanks. Yes i will try your PR asap.

JoergZeidler: Hi @jw-msft, 

thank you for providing PR #659 . Today i integrated the files into my adu-agent 1.1.0 package.
Unfortunately the fix didn't work. In my case the adu-agent stops in state "Invalid desired update action" in the root-key package download step during attempt to update from version "A" to "B"

Here my logs from the agent:

[log.txt](https://github.com/user-attachments/files/17493277/log.txt)

jw-msft: Thanks for trying it out, @JoergZeidler. I will try out the c2d msg from the log.txt to see what the issue was there as soon as I am able to.

JoergZeidler: @jw-msft:
imho, i found the issue in PR #659.

https://github.com/Azure/iot-hub-device-update/blob/50b6ef3a71ca1cdfef70739e1db8a65786ffe1e8/src/adu_workflow/src/agent_workflow.c#L441

should be changed to :
if (IsAducResultCodeFailure(result.ResultCode) || IsAducResultCodeFailure(rootkeyErc.ResultCode))

JoergZeidler: Hi @jw-msft,

I investigate some time to analyse whats going wrong. Imho, you can not check the "rootkeyErc" result here:

https://github.com/Azure/iot-hub-device-update/blob/50b6ef3a71ca1cdfef70739e1db8a65786ffe1e8/src/adu_workflow/src/agent_workflow.c#L441

I changed that line to the old implementation, and at least the normal update process will work again.

jw-msft: Thanks, @JoergZeidler, for sharing this finding--it is a correct understanding. I have pushed this commit: [34e6d5cd2f25934b9370fc351c9532e663b741ca](https://github.com/Azure/iot-hub-device-update/commit/34e6d5cd2f25934b9370fc351c9532e663b741ca) that also addresses: https://github.com/Azure/iot-hub-device-update/pull/659#discussion_r1785490268

JoergZeidler:
will be fixed by #659

PR #620, opened 2024-04-30
Adding pip authenticate to pipelines

nihemstr:
- Adds the PipAuthenticate@1 function for the checkout of the PIP artifacts to be used for various builds. 
- Removed unnecessary dependence on python2.7 which can cause flags
- Updated error warnings in install-deps.sh and install-deps.ps1 to use PipAuthenticate when working in pipelines or creating release artifacts

### DIFF
--- a/azurepipelines/e2e_test/templates/e2e_test_run.yaml
+++ b/azurepipelines/e2e_test/templates/e2e_test_run.yaml
@@ -11,7 +11,6 @@ steps:
       displayName: 'Pip Authenticate'
       inputs:
           artifactFeeds: 'adu-linux-client/ADUTestFeed'
-          onlyExtraIndex: true
 
     - script: |
           mkdir testresults
--- a/azurepipelines/e2e_test/templates/e2e_test_run.yaml
+++ b/azurepipelines/e2e_test/templates/e2e_test_run.yaml
@@ -7,6 +7,12 @@ parameters:
       type: string
 
 steps:
+    - task: PipAuthenticate@1
+      displayName: 'Pip Authenticate'
+      inputs:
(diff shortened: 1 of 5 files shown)


azurepipelines/e2e_test/templates/e2e_test_run.yaml
```diff
@@ -7,6 +7,12 @@ parameters:
       type: string
 
 steps:
+    - task: PipAuthenticate@1
+      displayName: 'Pip Authenticate'
+      inputs:
+          artifactFeeds: 'adu-linux-client/ADUTestFeed'
+          onlyExtraIndex: true
```
jw-msft, line 14: Why not making it the primary index URL?  I see `onlyAddExtraIndex` but not `onlyExtraIndex` in the pip authenticate v1 doc: https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/pip-authenticate-v1?view=azure-pipelines

azurepipelines/e2e_test/templates/e2e_vm_setup.yaml
```diff
@@ -26,6 +26,12 @@ parameters:
     - name: provisioning_type
       type: string
 steps:
+    - task: PipAuthenticate@1
+      displayName: 'Pip Authenticate'
+      inputs:
+          artifactFeeds: 'adu-linux-client/ADUTestFeed'
+          onlyExtraIndex: true
```
jw-msft, line 33: onlyAddExtraIndex here too?

jw-msft (changes requested):

jw-msft (approved):

PR #621, opened 2024-05-03
Re-Enabling native build pipelines for Ubuntu 20.04 and Ubuntu 22.04

nihemstr:
- Fixed native build pipeline issues
- Removed reliance on docker
- Changed dependency installation flow to prep for cache flows.

### DIFF
--- a/azurepipelines/build/native/adu-ubuntu-amd64-build.yml
+++ b/azurepipelines/build/native/adu-ubuntu-amd64-build.yml
@@ -85,8 +85,8 @@ extends:
         jobs:
           - job: BuildAduAgent_ubuntu2004
             displayName: "Build ADU Agent - Ubuntu 20.04 (amd64)"
-            timeoutInMinutes: 120
-            cancelTimeoutInMinutes: 120
+            timeoutInMinutes: 30
+            cancelTimeoutInMinutes: 30
             pool:
               name: aduc_1es_client_pool
               os: linux
@@ -98,8 +98,8 @@ extends:
 
           - job: BuildAduAgent_ubuntu2204
             displayName: "Build ADU Agent - Ubuntu 22.04 (amd64)"
-            timeoutInMinutes: 120
-            cancelTimeoutInMinutes: 120
+            timeoutInMinutes: 30
+            cancelTimeoutInMinutes: 30
             pool:
               name: aduc_1es_client_pool
               os: linux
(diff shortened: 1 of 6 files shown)


azurepipelines/build/native/adu-ubuntu-amd64-build.yml
```diff
@@ -83,28 +83,28 @@ extends:
     stages:
       - stage: Build
         jobs:
-          - job: BuildAduAgent_ubuntu2204
-            displayName: "Build ADU Agent - Ubuntu 22.04 (amd64)"
+          - job: BuildAduAgent_ubuntu2004
+            displayName: "Build ADU Agent - Ubuntu 20.04 (amd64)"
             timeoutInMinutes: 120
```
Nox-MSFT, line 88: Should we consider reducing the build timeout? The total build time should be less than 60 minutes (or 40 minutes).
Or, may be we should set timeout per job, inside the adu-native-build-steps.yml, if possible.

scripts/build.sh
```diff
@@ -46,6 +46,12 @@ install_adu=false
 work_folder=/tmp
 cmake_dir_path="${work_folder}/deviceupdate-cmake"
 
+#
+# Export the compiler settings in case VM is wonky
```
Nox-MSFT, line 50: What would be the common or expected errors when VM is "wonky"?

Nox-MSFT (approved):
Approved with some questions and suggestion

azurepipelines/build/native/adu-ubuntu-amd64-build.yml
jw-msft: Let's set it to 30 mins. The idea is if it takes longer than that, it can be considered a "failed build", possibly due to ADO infra issues OR it gives us incentive to make the builds as fast as possible, which is a good thing.

jw-msft (approved):

scripts/build.sh
nihemstr: CC/ CXX is not found seems to be an intermittent issue on some of these virtual machines.

The true fix would be to add a 1ES artifact that sets up and exports these variables in a programmatic way.

azurepipelines/build/native/adu-ubuntu-amd64-build.yml
nihemstr: This is a good bug that should be related to larger items not just the ubuntu build pipelines.

jw-msft (approved):

PR #656, opened 2024-09-12
Support for authentication with client certificates

D-r-P-3-p-p-3-r:
Added possibility to authenticate via client certificates (including configuration of CA certificates).

### DIFF
--- a/src/utils/config_utils/src/config_utils.c
+++ b/src/utils/config_utils/src/config_utils.c
@@ -75,68 +75,78 @@ static char* ADUC_AgentInfo_Read_X509_File(const char* const x509_file_path) {
     size_t bufferLength = 0;
     size_t bytesRead = 0;
     FILE* x509File = NULL;
-    
+    bool success = false;
+
     x509File = fopen(x509_file_path, "rb");
-    if (x509File)
+    if (x509File == NULL)
     {
-        if (fseek(x509File, 0, SEEK_END) == 0) {
-            ftellResult = ftell(x509File);
-            if (ftellResult != -1)
-            {
-                bufferLength = (size_t)ftellResult;
-                if (fseek(x509File, 0, SEEK_SET) == 0)
-                {
-                    buffer = malloc(bufferLength);
-                    if (buffer)
-                    {
-                        bytesRead = fread(buffer, 1, bufferLength, x509File);
-                        if (bytesRead < bufferLength)
-                        {
-                            if (feof(x509File))
-                            {
-                                Log_Error("Unexpected end of file while reading '%s' (bytesRead: %zu)!", x509_file_path, bytesRead);
-                            }
-                            else if (ferror(x509File)) {
-                                Log_Error("Error occurred while reading '%s' (bytesRead: %zu)!", x509_file_path, bytesRead);
-                            }
-                            else
-                            {
-                                Log_Error("Failed to read '%s' (bytesRead: %zu)!", x509_file_path, bytesRead);
-                            }
-                            free(buffer);
-                            buffer = NULL;
-                        }
-                    }
-                    else
-                    {
-                        Log_Error("Failed to get memory for buffer!");
-                    }
-                }
-                else
-                {
-                    Log_Error("Failed to seek begin of '%s'!", x509_file_path);
-                }
-            }
-            else
-            {
-                Log_Error("Failed to get file position ('%d')!", errno);
-            }
+        Log_Error("Failed to open '%s'!", x509_file_path);
+        goto done;
+    }
+
+    if (fseek(x509File, 0, SEEK_END) != 0)
+    {
+        Log_Error("Failed to seek end of '%s'!", x509_file_path);
+        goto done;
+    }
+
+    ftellResult = ftell(x509File);
+    if (ftellResult == -1)
+    {
+        Log_Error("Failed to get file position ('%d')!", errno);
+        goto done;
+    }
+
+    bufferLength = (size_t)ftellResult;
+    if (fseek(x509File, 0, SEEK_SET) != 0)
+    {
+        Log_Error("Failed to seek begin of '%s'!", x509_file_path);
+        goto done;
+    }
+
+    buffer = malloc(bufferLength);
+    if (buffer == NULL)
+    {
+        Log_Error("Failed to get memory for buffer!");
+        goto done;
+    }
+
+    bytesRead = fread(buffer, 1, bufferLength, x509File);
+    if (bytesRead < bufferLength)
+    {
+        if (feof(x509File))
+        {
+            Log_Error("Unexpected end of file while reading '%s' (bytesRead: %zu)!", x509_file_path, bytesRead);
+        }
+        else if (ferror(x509File)) {
+            Log_Error("Error occurred while reading '%s' (bytesRead: %zu)!", x509_file_path, bytesRead);
         }
         else
         {
-            Log_Error("Failed to seek end of '%s'!", x509_file_path);
+            Log_Error("Failed to read '%s' (bytesRead: %zu)!", x509_file_path, bytesRead);
         }
 
-        if(fclose (x509File) != 0) {
+        goto done;
+    }
+
+    success = true;
+
+done:
+    if (!success)
+    {
+        free(buffer);
+        buffer = NULL;
+    }
+
+    if (x509File != NULL)
+    {
+        if (fclose(x509File) != 0)
+        {
             Log_Warn("Failed to close '%s'!", x509_file_path);
         }
     }
-    else
-    {
-        Log_Error("Failed to open '%s'!", x509_file_path);
-    }
 
-    return(buffer);
+    return buffer;
 }
 
 /**
--- a/src/utils/config_utils/src/config_utils.c
+++ b/src/utils/config_utils/src/config_utils.c
@@ -14,13 +14,13 @@
 #include <aducpal/stdlib.h> // setenv
 #include <azure_c_shared_utility/crt_abstractions.h>
 #include <azure_c_shared_utility/strings_types.h>
+#include <errno.h>
 #include <parson.h>
 #include <parson_json_utils.h>
 #include <pthread.h>
 #include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
-#include <errno.h>
 
 static pthread_mutex_t s_config_mutex = PTHREAD_MUTEX_INITIALIZER;
 
@@ -68,8 +68,8 @@ static const char* INVALID_OR_MISSING_FIELD_ERROR_FMT = "Invalid json - '%s' mis
 
 static void ADUC_AgentInfo_Free(ADUC_AgentInfo* agent);
 
-
-static char* ADUC_AgentInfo_Read_X509_File(const char* const x509_file_path) {
+static char* ADUC_AgentInfo_Read_X509_File(const char* const x509_file_path)
+{
     char* buffer = NULL;
     long ftellResult = 0;
     size_t bufferLength = 0;
@@ -118,7 +118,8 @@ static char* ADUC_AgentInfo_Read_X509_File(const char* const x509_file_path) {
         {
             Log_Error("Unexpected end of file while reading '%s' (bytesRead: %zu)!", x509_file_path, bytesRead);
         }
-        else if (ferror(x509File)) {
+        else if (ferror(x509File))
+        {
             Log_Error("Error occurred while reading '%s' (bytesRead: %zu)!", x509_file_path, bytesRead);
         }
         else
@@ -191,8 +192,10 @@ static bool ADUC_AgentInfo_Init(ADUC_AgentInfo* agent, const JSON_Object* agent_
     connection_type = json_object_get_string(connection_source, CONFIG_CONNECTION_TYPE);
     connection_data = json_object_get_string(connection_source, CONFIG_CONNECTION_DATA);
     connection_x509_cert_file_path = json_object_get_string(connection_source, CONFIG_CONNECTION_X509_CERT_FILE_PATH);
-    connection_x509_private_key_file_path = json_object_get_string(connection_source, CONFIG_CONNECTION_X509_PRIVATE_KEY_FILE_PATH);
-    connection_x509_ca_cert_file_path = json_object_get_string(connection_source, CONFIG_CONNECTION_X509_CA_CERT_FILE_PATH);
+    connection_x509_private_key_file_path =
+        json_object_get_string(connection_source, CONFIG_CONNECTION_X509_PRIVATE_KEY_FILE_PATH);
+    connection_x509_ca_cert_file_path =
+        json_object_get_string(connection_source, CONFIG_CONNECTION_X509_CA_CERT_FILE_PATH);
 
     // As these fields are mandatory, if any of the fields doesn't exist, the agent will fail to be constructed.
     if (name == NULL || runas == NULL || connection_type == NULL || connection_data == NULL || manufacturer == NULL
@@ -221,8 +224,11 @@ static bool ADUC_AgentInfo_Init(ADUC_AgentInfo* agent, const JSON_Object* agent_
         goto done;
     }
 
-    if (connection_x509_cert_file_path || connection_x509_private_key_file_path || connection_x509_ca_cert_file_path) {
-        if(connection_x509_cert_file_path && connection_x509_private_key_file_path && connection_x509_ca_cert_file_path) {
+    if (connection_x509_cert_file_path || connection_x509_private_key_file_path || connection_x509_ca_cert_file_path)
+    {
+        if (connection_x509_cert_file_path && connection_x509_private_key_file_path
+            && connection_x509_ca_cert_file_path)
+        {
             agent->x509Cert = ADUC_AgentInfo_Read_X509_File(connection_x509_cert_file_path);
             if (!agent->x509Cert)
             {
@@ -238,8 +244,11 @@ static bool ADUC_AgentInfo_Init(ADUC_AgentInfo* agent, const JSON_Object* agent_
             {
                 goto done;
             }
-        } else {
-            Log_Error("Incomplete X509 client certificate configuration! Cert file path or private key file path is missing.");
+        }
+        else
+        {
+            Log_Error(
+                "Incomplete X509 client certificate configuration! Cert file path or private key file path is missing.");
         }
     }
 
--- a/src/utils/config_utils/src/config_utils.c
+++ b/src/utils/config_utils/src/config_utils.c
@@ -57,12 +57,62 @@ static const char* CONFIG_RUN_AS = "runas";
 static const char* CONFIG_CONNECTION_SOURCE = "connectionSource";
 static const char* CONFIG_CONNECTION_TYPE = "connectionType";
 static const char* CONFIG_CONNECTION_DATA = "connectionData";
+static const char* CONFIG_CONNECTION_X509_CERT_FILE_PATH = "connectionX509CertFilePath";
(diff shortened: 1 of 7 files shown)


D-r-P-3-p-p-3-r:
@microsoft-github-policy-service agree company="Seadex GmbH"

jw-msft:
Using EIS (edge identity service) to retrieve the x509 certs is already a supported way to use them for individual and group DPS enrollments. If I am interpreting correctly, what is the reasoning to support yet another way of doing it directly? Is it because you do not want to use EIS?

You can see here that we request the certificate from EIS unix domain server:
https://github.com/Azure/iot-hub-device-update/blob/18e98a95f5d77a608c2c224e50fe435814f4475b/src/utils/eis_utils/src/eis_utils.c#L621

Specifically, we are using https://azure.github.io/iot-identity-service/ with further details here: https://azure.github.io/iot-identity-service/api/identity-service.html

The reason device update agent uses EIS is for separating and offloading the concerns of securely handling the certicates where EIS can use a TPM or HSM, etc.

D-r-P-3-p-p-3-r:
> Using EIS (edge identity service) to retrieve the x509 certs is already a supported way to use them for individual and group DPS > enrollments. If I am interpreting correctly, what is the reasoning to support yet another way of doing it directly?

Yes.

> Is it because you do not want to use EIS?

Yes, because it is rather heavy-weight for our environment.
Building DUA and all its dependencies for an ARM-based system with a custom Linux (ptxdist-baked) was already very time-consuming. It will be very time-consuming to maintain in the future as well and we'll have to repeat this step in parts for DUA updates. Not only might DUA change, but so might its dependencies.

If we do the same with EIS it's even more binaries to build, configure, and maintain.

Side note:
It wasn't a lot of big issues, but DUA wasn't even 32-bit compatible out of the box.

Also, the system is constrained in memory and flash storage. It's not a tiny embedded controller. But adding a bunch of services just to do certificate-based authentication (which works without them) is not justified.
 
> The reason device update agent uses EIS is for separating and offloading the concerns of securely handling the certicates where > EIS can use a TPM or HSM, etc.

This is already implemented for our device and works just fine with OpenSSL out of the box. You don't point to a key file in the filesystem, but you use a "special" path to the key in the secure storage.

Thus, there is no benefit for us. But a lot of overhead. Especially, because the device software already uses the Azure IoT C SDK and uses the "direct way" to authenticate using the same key material. It works well and it's completely secure.

src/utils/config_utils/src/config_utils.c
```diff
@@ -57,13 +57,64 @@ static const char* CONFIG_RUN_AS = "runas";
...
+    char* buffer = NULL;
+    size_t bufferLength = 0;
+    size_t bytesRead = 0;
+    FILE* x509File = NULL;
+    
+    x509File = fopen(x509_file_path, "rb");
+    if (x509File)
+    {
+        if (fseek(x509File, 0, SEEK_END) == 0) {
+            bufferLength = (size_t)ftell(x509File);
```
jw-msft: Somewhat pedantic but how about handling when ftell returns -1 and Log_Error the errno. I suppose it would handle mostly the off chance when x509File is not a regular file (such as ESPIPE)

src/utils/config_utils/src/config_utils.c
```diff
@@ -57,13 +57,64 @@ static const char* CONFIG_RUN_AS = "runas";
...
+            if (fseek(x509File, 0, SEEK_SET) == 0)
+            {
+                buffer = malloc(bufferLength);
+                if (buffer)
+                {
+                    bytesRead = fread(buffer, 1, bufferLength, x509File);
+                    if (bytesRead < bufferLength)
+                    {
+                        Log_Error("Failed to read '%s'!", x509_file_path);
```
jw-msft: could eof and error be distinguished and also log the return value from bytesRead? Something like this:
```c
if (bytesRead < bufferLength)
{
    if (feof(x509File)) {
        Log_Error("Unexpected end of file while reading '%s', bytesRead: %zu", x509_file_path, bytesRead);
    } else if (ferror(x509File)) {
        Log_Error("Error occurred while reading '%s', bytesRead: %zu", x509_file_path, bytesRead);
    }
    free(buffer);
    buffer = NULL;
}
```

jw-msft (commented):
Looks good. A couple of minor comments.

D-r-P-3-p-p-3-r:
Updated the PR with the suggest changes.

src/utils/config_utils/src/config_utils.c
D-r-P-3-p-p-3-r: done

src/utils/config_utils/src/config_utils.c
D-r-P-3-p-p-3-r: done

src/utils/config_utils/src/config_utils.c
```diff
@@ -130,6 +211,28 @@ static bool ADUC_AgentInfo_Init(ADUC_AgentInfo* agent, const JSON_Object* agent_
         goto done;
     }
 
+    if (connection_x509_cert_file_path || connection_x509_private_key_file_path || connection_x509_ca_cert_file_path) {
+        if(connection_x509_cert_file_path && connection_x509_private_key_file_path && connection_x509_ca_cert_file_path) {
+            agent->x509Cert = ADUC_AgentInfo_Read_X509_File(connection_x509_cert_file_path);
+            if (!agent->x509Cert)
```
jw-msft, line 236: the code base uses explicit conditions regarding check for NULL values, so please change to:
```c
if (agent->x509Cert == NULL)
```

The same for lines 221 and 226.

src/utils/config_utils/src/config_utils.c
```diff
@@ -57,13 +58,87 @@ static const char* CONFIG_RUN_AS = "runas";
...
+
+static char* ADUC_AgentInfo_Read_X509_File(const char* const x509_file_path) {
+    char* buffer = NULL;
+    long ftellResult = 0;
+    size_t bufferLength = 0;
+    size_t bytesRead = 0;
+    FILE* x509File = NULL;
+    
+    x509File = fopen(x509_file_path, "rb");
+    if (x509File)
```
jw-msft: use explicit checks for NULL / non-NULL:
```c
if (x509File != NULL)
```

src/utils/config_utils/src/config_utils.c
```diff
@@ -57,13 +58,87 @@ static const char* CONFIG_RUN_AS = "runas";
...
+                bufferLength = (size_t)ftellResult;
+                if (fseek(x509File, 0, SEEK_SET) == 0)
+                {
+                    buffer = malloc(bufferLength);
+                    if (buffer)
+                    {
+                        bytesRead = fread(buffer, 1, bufferLength, x509File);
+                        if (bytesRead < bufferLength)
+                        {
+                            if (feof(x509File))
```
jw-msft: Line 95 here has an if-statement that is 7 levels deep.
Let's keep the nesting of if-statements shallow by using the following pattern that:
- guarantees usually just 1-level of nesting, and usually at most 2-3 nesting depth in practice
- Centralizes all clean-up code after 'done' label that in general avoids doing so in multiple nested error branches (not the case here but in general it lowers the cognitive load), similar in function to Zig or Odin's defer statement (or c++ RAII) but can control exactly when it happens:
```c
bool success = false;
...
if (x509File == NULL)
{
    Log_Error("Failed to open '%s'!", x509_file_path);
    goto done;
}

if (fseek(x509File, 0, SEEK_END) != 0)
{
    Log_Error("Failed to seek end of '%s'!", x509_file_path);
    goto done;
}

...
...
    success = true;
done:
    if (!success)
    {
        free(buffer);
        buffer = NULL;
    }

    if (x509File != NULL)
    {
        fclose(x509File);
    }

    return buffer;
}
```

src/utils/config_utils/src/config_utils.c
```diff
@@ -57,13 +58,87 @@ static const char* CONFIG_RUN_AS = "runas";
...
+    char* buffer = NULL;
+    long ftellResult = 0;
+    size_t bufferLength = 0;
+    size_t bytesRead = 0;
+    FILE* x509File = NULL;
+    
+    x509File = fopen(x509_file_path, "rb");
+    if (x509File)
+    {
+        if (fseek(x509File, 0, SEEK_END) == 0) {
```
jw-msft: Style consistency issue: Opening brace should be on next line

Touch every file in the current set of commits, `git add` them to git staging, and then run `scripts/clang-format.sh`. see: https://github.com/Azure/iot-hub-device-update/blob/0f041bfc6cd25593ed61d58e4113b2a4dad7394f/scripts/clang-format.sh#L10

Longer-term, we need to add clang-format.sh (and some other things) to the git pre-commit hook.  Currently, the existing git pre-commit hook only runs sh-format.sh (https://github.com/Azure/iot-hub-device-update/blob/0f041bfc6cd25593ed61d58e4113b2a4dad7394f/scripts/githooks/pre-commit.sh#L29)

src/utils/config_utils/src/config_utils.c
```diff
@@ -57,13 +58,87 @@ static const char* CONFIG_RUN_AS = "runas";
...
+        if(fclose (x509File) != 0) {
+            Log_Warn("Failed to close '%s'!", x509_file_path);
+        }
+    }
+    else
+    {
+        Log_Error("Failed to open '%s'!", x509_file_path);
+    }
+
+    return(buffer);
```
jw-msft: style: no parens for return

src/utils/config_utils/src/config_utils.c
```diff
@@ -57,13 +58,87 @@ static const char* CONFIG_RUN_AS = "runas";
...
+            {
+                Log_Error("Failed to get file position ('%d')!", errno);
+            }
+        }
+        else
+        {
+            Log_Error("Failed to seek end of '%s'!", x509_file_path);
+        }
+
+        if(fclose (x509File) != 0) {
```
jw-msft: style: no space between function name and arg list (should be caught by clang-format)

src/utils/config_utils/src/config_utils.c
```diff
@@ -57,13 +58,87 @@ static const char* CONFIG_RUN_AS = "runas";
...
+                    if (buffer)
+                    {
+                        bytesRead = fread(buffer, 1, bufferLength, x509File);
+                        if (bytesRead < bufferLength)
+                        {
+                            if (feof(x509File))
+                            {
+                                Log_Error("Unexpected end of file while reading '%s' (bytesRead: %zu)!", x509_file_path, bytesRead);
+                            }
+                            else if (ferror(x509File)) {
```
jw-msft: paren on next line

src/utils/config_utils/src/config_utils.c
```diff
@@ -57,13 +58,87 @@ static const char* CONFIG_RUN_AS = "runas";
...
+                Log_Error("Failed to get file position ('%d')!", errno);
+            }
+        }
+        else
+        {
+            Log_Error("Failed to seek end of '%s'!", x509_file_path);
+        }
+
+        if(fclose (x509File) != 0) {
+            Log_Warn("Failed to close '%s'!", x509_file_path);
```
jw-msft: omit. no need to log when fclose fails as it's not really actionable and the log trace might not even be able to flush to disk in that case (things are in a pretty bad state)

src/utils/config_utils/src/config_utils.c
jw-msft: would prefer this to be in centralized cleanup area, such as after 'done:' label (along with free(buffer) if in error condition).

src/utils/config_utils/src/config_utils.c
```diff
@@ -57,13 +58,87 @@ static const char* CONFIG_RUN_AS = "runas";
...
+                                Log_Error("Unexpected end of file while reading '%s' (bytesRead: %zu)!", x509_file_path, bytesRead);
+                            }
+                            else if (ferror(x509File)) {
+                                Log_Error("Error occurred while reading '%s' (bytesRead: %zu)!", x509_file_path, bytesRead);
+                            }
+                            else
+                            {
+                                Log_Error("Failed to read '%s' (bytesRead: %zu)!", x509_file_path, bytesRead);
+                            }
+                            free(buffer);
```
jw-msft: for future maintainability, this should be done unconditionally at the end, if in "not succeeded" state.

jw-msft (commented):
mostly style comments, some of which can be fixed by running clang-format.

D-r-P-3-p-p-3-r:
I made all the changes as requested except removing the result check on fclose. I can't bring myself to do that. It doesn't feel right to remove an error check and the diagnostics even if the chance of them being emitted are small.
If you feel the need to do so, go ahead... :)

ZhenyuWang-CVX:
@D-r-P-3-p-p-3-r , I am trying out this branch per suggestion from MSFT's ADU team. My goal is to configure the ADUC_Agent to use X509 certificates to communication with the IoT Hub.

I am able to pick your this branch and build. In my testing, the code keeps failing at the function ADUC_DeviceClient_Create() in iothub_communication_manager.c. where it tries to call ClientHandle_CreateFromConnectionString() to create a connection string. 

I think the issue is due to the syntax in my config file (du-config.json) is not right. I am wondering whether you have encountered similar issue and would you mind to share your configuration file template?

D-r-P-3-p-p-3-r:
@ZhenyuWang-CVX 

This is how our working template looks like:

```
{
	"schemaVersion": "1.1",
	"aduShellTrustedUsers": [
		"adu",
		"do"
	],
	"iotHubProtocol": "mqtt",
	"manufacturer": "<MANUFACTURER>",
	"model": "<MODEL>",
	"agents": [
		{
			"name": "aduagent",
			"runas": "adu",
			"connectionSource": {
				"connectionType": "X509",
				"connectionData": "HostName=<IOT_HUB_DNS_NAME>;DeviceId=<DEVICE_ID>;x509=true",
				"connectionX509CertFilePath": "<PATH_TO_CLIENT_CERT>",
				"connectionX509PrivateKeyFilePath": "<PATH_TO_CLIENT_PRIVATE_KEY>",
				"connectionX509CaCertFilePath": "<PATH_TO_FILE_CONTAINING_APPENDED_MSFT_CA_CERTS"
			},
			"manufacturer": "<MANUFACTURER>",
			"model": "<MODEL>"
		}
	]
}
```


Maybe you are missing the x509 part?

ZhenyuWang-CVX:
@D-r-P-3-p-p-3-r  thank you very for your sharing. The missing part apparently is "x509=true
".
"connectionData": "HostName=<IOT_HUB_DNS_NAME>;DeviceId=<DEVICE_ID>;**_x509=true_**",

D-r-P-3-p-p-3-r:
@ZhenyuWang-CVX 
Glad it works for you now!

src/utils/config_utils/src/config_utils.c
```diff
@@ -130,6 +224,34 @@ static bool ADUC_AgentInfo_Init(ADUC_AgentInfo* agent, const JSON_Object* agent_
...
+            }
+            agent->x509CaCert = ADUC_AgentInfo_Read_X509_File(connection_x509_ca_cert_file_path);
+            if (!agent->x509CaCert)
+            {
+                goto done;
+            }
+        }
+        else
+        {
+            Log_Error(
```
jw-msft, line 267: should `goto done` here also so it does not proceed and possibly set success = true

smithsonarrey-chevron:
Hi @D-r-P-3-p-p-3-r, I tried this branch out and it seems like a great alternative to using EIS! I think we are trying to solve the same problem. I'm curious, does your device software transmit data to the IoT Hub? If so, is it the same IoT Hub that you have ADU deployed to? 

In our case, our device software experienced many telemetry failures after pushing data and to same IoT Hub that ADU is deployed. However, our ADU Agent appeared to be unaffected by whatever telemetry was sent by our software. 

Is this something that you've experienced?

D-r-P-3-p-p-3-r:
@smithsonarrey-chevron: Are you using the same identity for ADU and telemetry? Or do you use a module identity for ADU?

smithsonarrey-chevron:
@D-r-P-3-p-p-3-r , it's the same identity

D-r-P-3-p-p-3-r:
@smithsonarrey-chevron Figured :)
That's a bad idea in my experience as the two connections to the same identity interfere with each other.
I'd bet that if you push your ADU connection to a module identity then your problems will go away.

smithsonarrey-chevron:
@D-r-P-3-p-p-3-r, we don't have modules on our devices since it is very lightweight and battery powered. Do you have any specific reason why it wouldn't work? In our discussions with microsoft, they stated that that there **shouldn't** be a conflict.

D-r-P-3-p-p-3-r:
AFAIR we had issues with connection losses. If you use ADU _and_ send telemetry data or work with twins, don't you need a second IoT hub connection anyway?

smithsonarrey-chevron:
That was actually the architecture we had previously. Didn't run into any issues with that but we'd prefer to only use one connection if it's possible. Looks like it won't be easy... Thank you, I appreciate your responses.

smithsonarrey-chevron:
Yeah, we're going to have to abandon the single device identity approach. Bummer, especially since we wanted to auto-provision to the IoT Hub using Device Provisioning Service. It doesn't seem to support module identities yet. I imagine you're either writing a script to provision your module identities or provisioning them manually?

D-r-P-3-p-p-3-r:
The device does the provisioning (of the main identity) and the module identities are created on the Azure IoT Hub/server side via backend logic.

smithsonarrey-chevron:
> the module identities are created on the Azure IoT Hub/server side via backend logic

Do you have any guidance on this? This is just an event-driven app or function, right?

Are you also using the same certs for the ADU module identity and your device identity? I imagine that this would be the simplest to manage.

Again, we really appreciate your support on this! You've been a big help.

D-r-P-3-p-p-3-r:
> > the module identities are created on the Azure IoT Hub/server side via backend logic
> 
> Do you have any guidance on this? This is just an event-driven app or function, right?

Frankly, idk. It's a different team developing the cloud code. I'm working on the embedded side.

> Are you also using the same certs for the ADU module identity and your device identity? I imagine that this would be the simplest to manage.

Yes. It's the very same key/certificate that is used for all connections in our scenario.

smithsonarrey-chevron:
That's fair, thank you for sharing :)

D-r-P-3-p-p-3-r:
I added another commit adding PKCS11 support for the private key.

In order to use the feature
* add _opensslEngine_ in the _connectionSource_ section of the configuration (e.g. "pkcs11"), and
* set the PKCS11 URI  of the private key instead of a file path in _connectionX509PrivateKeyFilePath_.

src/utils/config_utils/src/config_utils.c
```diff
@@ -57,13 +58,99 @@ static const char* CONFIG_RUN_AS = "runas";
...
+static const char* CONFIG_CONNECTION_OPENSSL_ENGINE = "opensslEngine";
+static const char* CONFIG_CONNECTION_X509_CA_CERT_FILE_PATH = "connectionX509CaCertFilePath";
 static const char* CONFIG_ADDITIONAL_DEVICE_PROPERTIES = "additionalDeviceProperties";
 static const char* CONFIG_AGENTS = "agents";
 
 static const char* INVALID_OR_MISSING_FIELD_ERROR_FMT = "Invalid json - '%s' missing or incorrect";
 
 static void ADUC_AgentInfo_Free(ADUC_AgentInfo* agent);
 
+static char* ADUC_AgentInfo_Read_X509_File(const char* const x509_file_path)
```
jw-msft, line 72: this is not specific to x509 files so should be put in a file utils static lib under src/utils.

jw-msft (approved):

Issue #625, opened 2024-05-13
Update status changes to "In Progress" sometime after "Succeeded".


mahdighorbanpourptw:
I have an update script which is a copy of the sample scripts and some changes to adapt it to our needs. The update works perfectly fine. The state changes from not running, in progress, succeeded. Then after a while, it changes back to In Progress. I looked at the update agent logs and there, the condition check works perfectly fine, and the update is not triggered twice! Note that, when I restart the agent, it immediately shows "Succeeded". Then the same issue, after some time, goes back to in progress. 
having said, the IotHub portal, updates, update-group overview, showing inconsistent data. The first two devices, are already updated successfully, which on the upper side info box, shows correctly. But in the lower side, all of them are shown "In Progress". The third device in the list is even turned off! 
![image](https://github.com/Azure/azure-sdk-for-net/assets/97948692/c3913de5-ccca-4ff1-858a-cd9d7add4faa)

You could also see that, in the twin, it says applied.

![image](https://github.com/Azure/azure-sdk-for-net/assets/97948692/e5f70aa6-f017-40b5-af72-47affed26546)

### Expected behavior

The first two devices should keep showing the state "Succeeded". And the last one, should be "Not Running"

### Actual behavior

All the devices are shown "In Progress". 

### Reproduction Steps

Use the example scripts to create an update. Assign it to the devices.

### Environment

Ubuntu 22.04
IotEdge 1.4.39
DeviceIpdate-Agent 1.1.0

josephmsft:
Is this "InProgress" state persisting continuously for the online devices? What is possibly happening is that if the device remains connected for over an hour, the IoT Hub connection can be refreshed and the agent will receive the ADM configuration again and go into the deployment workflow and report InProgress, but once in the workflow it should see it has the update installed already and report Success again. I would not expect it to stay "InProgress" for an extended period of time.

Was the offline agent ever online to receive the deployment? Or has it been offline the entire time?

mahdighorbanpourptw:
This is the last shell log file from a device that is showing in progress few minutes after success.

```
2024-05-14T04:40:27.1399Z 319711[319711] [I] #   [ShowChildProcessLogs:234]
2024-05-14T04:40:27.1399Z 319711[319711] [I] #  [2024/05/14:044027] [I] (2)List workfolder directory content: total 40468 [ShowChildProcessLogs:234]
[...]

PR #659, opened 2024-09-26
Fixed bug reporting in progress out of band, re-introducing InProgress workflow step

nihemstr:


src/adu_workflow/src/agent_workflow.c
```diff
@@ -1327,6 +1336,36 @@ void ADUC_Workflow_MethodCall_Idle(ADUC_WorkflowData* workflowData)
...
+{
+    ADUC_WorkflowData* workflowData = methodCallData->WorkflowData;
+
+    ADUC_Result result = { .ResultCode = ADUC_Result_Success , .ExtendedResultCode = 0 };
+    Log_Info("Workflow step: ProcessDeployment");
+
+    //
+    // Shouldn't have to handle anything else here. WorkflowData already made?
+    //
+
```
jw-msft, line 1355: 2 newlines in a row. touch, stage, and run clang-format?

src/agent/src/main.c
```diff
@@ -831,7 +831,7 @@ void ShutdownAgent()
  */
 void OnShutdownSignal(int sig)
 {
-    Log_Warn("Shutdown signal detected: %d", sig);
+    Log_Warn("Shutdown signal detected: %s", sig);
```
jw-msft: the fix here seems to be regressing a previous fix. It should be %d to avoid undefined behavior (including segfault)

jw-msft (changes requested):
I am blocking this PR because need to revert the change to line 834 on main.c which was previously fixing the undefined behavior / segfault.

src/agent/src/main.c
jw-msft: I've addressed this in https://github.com/Azure/iot-hub-device-update/commit/34e6d5cd2f25934b9370fc351c9532e663b741ca

jw-msft (approved):

jw-msft (approved):

Nox-MSFT (approved):

ortogonal:
Just for information, I tested this PR and had some issues. See my comment here, https://github.com/Azure/iot-hub-device-update/issues/625#issuecomment-2435455883

RichardHubertPtw:
Thanks for making some progress here! Please let us know if we can assist with tests or is there an alternative approach in the works? If so, please inform in the original issue ASAP. We are getting desperate :-)

PR #684, opened 2025-01-15
Update eis_utils, increase AIS response buffer size, add more error logs, and change ADU Snap version to 0.2

Nox-MSFT:
- Increase AIS response buffer size
- Add more error logs
- Change ADU Snap version to 0.2

### DIFF
--- a/src/utils/eis_utils/src/eis_coms.c
+++ b/src/utils/eis_utils/src/eis_coms.c
@@ -8,6 +8,7 @@
 
 #include "eis_coms.h"
 
+#include <aduc/logging.h>
 #include <aduc/string_c_utils.h>
 #include <azure_c_shared_utility/azure_base64.h>
 #include <azure_c_shared_utility/buffer_.h>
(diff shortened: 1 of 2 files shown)


src/utils/eis_utils/src/eis_coms.c
```diff
@@ -152,7 +153,7 @@ typedef struct tagEIS_HTTP_WORKLOAD_CONTEXT
 /**
  * @brief Maximum amount of bytes for any EIS response
  */
-#define EIS_RESP_SIZE_MAX 4096
+#define EIS_RESP_SIZE_MAX 8192
```
jw-msft, line 156: Is this the true max, or should x2 for safety factor?

jw-msft (approved):
Approved with comment

src/utils/eis_utils/src/eis_coms.c
Nox-MSFT: x2 for safety factor at the moment.

PR #689, opened 2025-01-20
Public GitHub Actions for build and test

AndreRicardo-Zoetis:
As suggested in discussion "Public GitHub Actions builds"
https://github.com/Azure/iot-hub-device-update/discussions/688

Example run:

https://github.com/AndreRicardo-Zoetis/iot-hub-device-update/actions/runs/12871357147

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -1,0 +1,25 @@
+name: CI
+on:
+  push:
+  workflow_dispatch:
+jobs:
+  container-job:
+    strategy:
+      matrix:
+        os: [debian:11, debian:10, ubuntu:22.04, ubuntu:20.04]
+    runs-on: ubuntu-latest
+    container:
+      image: ${{ matrix.os }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Setup
+        run: |
+          apt-get update
+          scripts/install-deps.sh --install-aduc-deps --install-do --install-cmake --install-shellcheck
+      - name: Build tests
+        run: scripts/build.sh --clean --build-unit-tests
+      - name: Run tests
+        run: |
+          cd out
+          ctest
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -13,12 +13,12 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      - name: prereqs
-        run: |
-          azurepipelines/e2e_test/packer/deltagen/scripts/install-prereqs.sh
-      - name: Setup
+      - name: Install prereqs
         run: |
           apt-get update
+          azurepipelines/e2e_test/packer/deltagen/scripts/install-prereqs.sh
+      - name: Install dependencies
+        run: |
           scripts/install-deps.sh --install-aduc-deps --install-do --install-cmake --install-shellcheck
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,30 @@
...
+        uses: actions/checkout@v4
+      - name: Install prereqs
+        run: |
+          apt-get update
+          # Same fix as azurepipelines/e2e_test/packer/deltagen/scripts/install-prereqs.sh for tzdata
+          DEBIAN_FRONTEND=noninteractive apt-get install -y tzdata
+          dpkg-reconfigure -f noninteractive tzdata
+      - name: Install dependencies
+        run: |
+          scripts/install-deps.sh --install-aduc-deps --install-do --install-cmake --install-shellcheck
```
jw-msft, line 23: any reason not doing -a instead?

jw-msft (dismissed):
looks good. thanks for doing this.

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,30 @@
...
+    runs-on: ubuntu-latest
+    container:
+      image: ${{ matrix.os }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Install prereqs
+        run: |
+          apt-get update
+          # Same fix as azurepipelines/e2e_test/packer/deltagen/scripts/install-prereqs.sh for tzdata
```
jw-msft: Could you remove this comment please as the az pipelines may be moved in near future?

AndreRicardo-Zoetis: Removed.

.github/workflows/docker-build.yml
AndreRicardo-Zoetis: Yes `-a` requires the install of `--install-githooks` which fails (see log below). 

I suspect would need to also checkout of `.git` folder and a `git config --global --add safe.directory "$(pwd)"` for the hooks to work. None of these are needed in CI.

```sh
Installing githooks...
Unable to determine git root.
Error: Process completed with exit code 1.
```
Example build with the error shown:
https://github.com/AndreRicardo-Zoetis/iot-hub-device-update/actions/runs/13541395990/job/37843105763#step:5:2124

Source of the error:
https://github.com/Azure/iot-hub-device-update/blob/develop/scripts/install-deps.sh#L159

I wonder if `-a` should include the git hooks because this dependency is more of a developer convenience, for CI builds it does not seem relevant.

jw-msft: good point. the git rev-parse evidently not working, so would need to try the safe-directory you mentioned. Already have --install-githooks which would still be a part of -a. I think we can add , say, a --install-ghactions for github action pipelines that encapsulates what you've done here.

jw-msft (approved):

AndreRicardo-Zoetis:
Not sure why the build on Debian 10 failed.

```sh
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  CMake 3.16 or higher is required.  You are running version 3.13.4
```
https://github.com/Azure/iot-hub-device-update/actions/runs/13556459435/job/37891642088#step:5:2665

`--install-cmake` is passed and the version on install-deps.sh is `supported_cmake_version='3.23.2'`

Seems catch2 is still using the system cmake version to compile.

PR #696, opened 2025-02-19
Fix message processing context init, usages, and cleanup.

Nox-MSFT:


src/utils/d2c_messaging/src/d2c_messaging.c
```diff
@@ -427,15 +460,17 @@ bool ADUC_D2C_Messaging_Init()
         memset(&s_pendingMessageStore, 0, sizeof(s_pendingMessageStore));
         for (i = 0; i < ADUC_D2C_Message_Type_Max; i++)
         {
-            s_messageProcessingContext[i].type = i;
-            s_messageProcessingContext[i].transportFunc = ADUC_D2C_Default_Message_Transport_Function;
-            s_messageProcessingContext[i].retryStrategy = &g_defaultRetryStrategy;
-            int res = pthread_mutex_init(&s_messageProcessingContext[i].mutex, NULL);
+             int res = pthread_mutex_init(&s_messageProcessingContext[i].mutex, NULL);
```
jw-msft, line 463: seems to be a leading space.

src/utils/d2c_messaging/src/d2c_messaging.c
```diff
@@ -456,7 +491,7 @@ void ADUC_D2C_Messaging_Uninit()
     if (s_core_initialized)
     {
         // Cancel pending messages
-        for (int i = 0; i < ADUC_D2C_Message_Type_Max; i++)
+        for (int i = 0; i < ADUC_D2C_Message_Type_Max && s_messageProcessingContext[i].initialized; i++)
```
jw-msft, line 494: maybe instead of stopping loop when encountering initialized==false, just skip in loop body for more robustness in case there is type in between that is not initialized but one after it is initialized.
i.e.
```c
for (...)
{
    if (false == s_messageProcessingContext[i].initialized)
    {
        continue;
    }
    ...
}
```

src/utils/d2c_messaging/src/d2c_messaging.c
```diff
@@ -546,6 +581,11 @@ bool ADUC_D2C_Message_SendAsync(
  */
 void ADUC_D2C_Messaging_Set_Transport(ADUC_D2C_Message_Type type, ADUC_D2C_MESSAGE_TRANSPORT_FUNCTION transportFunc)
 {
+    if (!s_messageProcessingContext[type].initialized)
+    {
+        Log_Error("Message processing context (0x%x) is not initialized.", &s_messageProcessingContext[type]);
+        return;
```
jw-msft, line 587: Can we do at least debug assertion for ADUC_D2C_Message_Type range check so it does not go out of bounds if for some reason random integer got cast to the enum?

src/utils/d2c_messaging/src/d2c_messaging.c
```diff
@@ -456,7 +491,7 @@ void ADUC_D2C_Messaging_Uninit()
     if (s_core_initialized)
     {
         // Cancel pending messages
-        for (int i = 0; i < ADUC_D2C_Message_Type_Max; i++)
+        for (int i = 0; i < ADUC_D2C_Message_Type_Max && s_messageProcessingContext[i].initialized; i++)
         {
             pthread_mutex_lock(&s_messageProcessingContext[i].mutex);
```
jw-msft, line 496: Should this be locking s_pendingMessageStore[i].mutex and lock again on line 502 for s_messageProcessingContext[i].mutex -- the idea being separate mutex for each granular data array?

jw-msft: Is s_pendingMessageStoreMutex being used as transaction lock? Instead, if possible, it should only be locking access to s_pendingMessageStore array. For s_core_initialized, lock can be avoided if can guarantee init and unint are called on main thread only, or else have yet another lock just for accessing s_core_initialized.

src/utils/d2c_messaging/src/d2c_messaging.c
```diff
@@ -456,7 +491,7 @@ void ADUC_D2C_Messaging_Uninit()
...
             if (s_pendingMessageStore[i].content != NULL)
```
jw-msft, line 497: I do not see where s_pendingMessageStore[i].content is set? Basically, it seems this will always be NULL (and if-stmt always will be false).  Can it be removed (along with its mutex)?

jw-msft: Maybe larger scope than this PR, but having mutex per s_messagePorcessingContext element (ADUC_D2C_Message_Type)seems excessive since there is only one worker thread processing these. We could increase in the future, but I doubt there will be multiple worker threads for any one type so probably sufficient to just have a single read-writer lock for accessing all the types.

src/utils/d2c_messaging/src/d2c_messaging.c
```diff
@@ -205,6 +205,19 @@ static void DefaultIoTHubSendReportedStateCompletedCallback(int http_status_code
...
+
+    if (message_processing_context == NULL)
+    {
+        Log_Error("context is NULL");
+        return;
+    }
+
+    if (!message_processing_context->initialized)
+    {
+        Log_Warn("Message processing context (0x%x) is not initialized.", message_processing_context);
```
jw-msft, line 217: Seems like should be Log_Error for more visibility since this should not be expected, correct?
Consider adding Debug assert here so debug build will break into debugger if attached.

jw-msft (approved):
Approved with some comments

jw-msft (approved):
Approved with some comments for possible future improvement

PR #701, opened 2025-02-25
Fix an issue in the Script Handler where the primary script file permissions were set incorrectly

Nox-MSFT:
**Cause**

The script_tasks incorrectly removed the read-only access for 'other' user, causing subsequent file hash calculations to fail and triggering multiple re-downloads of the same file.

### DIFF
--- a/src/adu-shell/src/script_tasks.cpp
+++ b/src/adu-shell/src/script_tasks.cpp
@@ -49,7 +49,7 @@ ADUShellTaskResult Execute(const ADUShell_LaunchArguments& launchArgs)
     bool filePermissionsChanged = false;
     bool statOk = stat(path, &st) == 0;
 
-        // Note: delivery optimization will create the file with do:do ownership
+    // Note: delivery optimization will create the file with do:do ownership
     // Here, we must allow the read access by the owner group and others (S_IRGRP) users (S_IROTH).
     // Otherwise, the script handler will fail to calculate the hash of the script.
     int mode = S_IRWXU | S_IRGRP | S_IXGRP | S_IROTH;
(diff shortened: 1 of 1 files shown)


AndreRicardo-Zoetis:
Does this bug trigger `ADUC_ERC_VALIDATION_FILE_HASH_INVALID_HASH` ?

src/adu-shell/src/script_tasks.cpp
```diff
@@ -48,7 +48,12 @@ ADUShellTaskResult Execute(const ADUShell_LaunchArguments& launchArgs)
     struct stat st = {};
     bool filePermissionsChanged = false;
     bool statOk = stat(path, &st) == 0;
-    int mode = S_IRWXU | S_IRGRP | S_IXGRP;
+
+    // Note: delivery optimization will create the file with do:do ownership
+    // Here, we must allow the read access by the owner group and others (S_IRGRP) users (S_IROTH).
+    // Otherwise, the script handler will fail to calculate the hash of the script.
+    int mode = S_IRWXU | S_IRGRP | S_IXGRP | S_IROTH;
```
jw-msft, line 55: Instead of this, perhaps on the repro machine `adu` user is not in `do` group, or DO Agent did not `chown` the file into `do` group?

jw-msft (changes requested):

Nox-MSFT:
Will actually change file ownership instead.

AndreRicardo-Zoetis:
@eshashah-msft / @Nox-MSFT / @jw-msft does this bug mean we will see `00100003` error code?

```json
"lastInstallResult": {
      "extendedResultCode": 0,
      "extendedResultCodes": "00100003,F0000FFF",
      "resultCode": 0,
      "resultDetails": "Need to install: 1.7.2.",
      "stepResults": {
          "step_0": {
              "extendedResultCodes": "00100003",
              "resultCode": 0,
              "resultDetails": "Need to install: 1.7.2."
          },      
```

PR #683, opened 2025-01-09
Migrate catch2 v2 to v3

AndreRicardo-Zoetis:
Hello, would you consider upgrading Catch to version 3?

We would like to use newer functionality in Catch2 such as SKIP.
https://catch2-temp.readthedocs.io/en/latest/skipping-passing-failing.html

Also noticed that in Catch2 version 3 there is no need to define all those empty `main.cpp`.

## Changes

Followed the guidelines from:
https://catch2-temp.readthedocs.io/en/latest/migrate-v2-to-v3.html

1. Change Catch2 to Catch2WithMain where `main.cpp` was used

2. `[!hide]` tag has been removed 
https://github.com/catchorg/Catch2/blob/914aeecfe23b1e16af6ea675a4fb5dbd5a5b8d0a/docs/release-notes.md?plain=1#L505

3. "The `Contains` string matcher was renamed to `ContainsSubstring`."
https://github.com/catchorg/Catch2/blob/devel/docs/migrate-v2-to-v3.md#things-that-can-break-during-porting

I can tidy up the PR further and remove all the no longer needed `main.cpp` files, but before doing so would like to have your guidance.

PS: I do have other changes to be able to run the tests as GitHub Action but was able to run the tests successfully in
https://github.com/AndreRicardo-Zoetis/iot-hub-device-update/actions/runs/12689720900/job/35369056358

PR #702, opened 2025-02-27
Fix builds failing for catch2 v3.8 by using specified cmake version

AndreRicardo-Zoetis:
Fix 
```sh
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  CMake 3.16 or higher is required.  You are running version 3.13.4
```

As mentioned on https://github.com/Azure/iot-hub-device-update/pull/689#issuecomment-2687481888

The combination of PR [Migrate catch2 v2 to v3 #683](https://github.com/Azure/iot-hub-device-update/pull/683) and [Public GitHub Actions for build and test #689](https://github.com/Azure/iot-hub-device-update/pull/689) sees that Debian 10 is using the system version of cmake 3.13.4 that is lower than the required by catch2 v3.8, which is cmake 3.16 https://github.com/catchorg/Catch2/blob/v3.8.0/CMakeLists.txt#L1

Tested on
https://github.com/AndreRicardo-Zoetis/iot-hub-device-update/actions/runs/13564853498/job/37915573097

Also tested locally with 
```sh
scripts/install-deps.sh --install-aduc-deps --install-do --install-shellcheck
```
`--install-aduc-deps` sets `install_catch2=true`, so it also calls the install catch2.

PS: also fixed what I believe is a typo on the name of a variable for shellcheck version.

### DIFF
--- a/scripts/install-deps.sh
+++ b/scripts/install-deps.sh
@@ -66,6 +66,7 @@ cmake_force_source=false
 cmake_prefix="$work_folder"
 cmake_installer_dir=""
 cmake_dir_symlink="/tmp/deviceupdate-cmake"
+cmake_bin="cmake"
 
 install_shellcheck=false
 supported_shellcheck_version='0.8.0'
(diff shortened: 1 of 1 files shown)


jw-msft:
thanks, I'm doing a run of this in gh codespace

jw-msft (approved):

PR #704, opened 2025-02-27
Ensure the downloaded script file has the correct ownership

Nox-MSFT:
Fix the Script Handler to ensure the script file has the correct ownership.
This to ensure that only 'adu' user and 'adu' group has access to the script file.

### DIFF
--- a/src/adu-shell/src/script_tasks.cpp
+++ b/src/adu-shell/src/script_tasks.cpp
@@ -51,19 +51,6 @@ ADUShellTaskResult Execute(const ADUShell_LaunchArguments& launchArgs)
     int mode = S_IRWXU | S_IRGRP | S_IXGRP;
     if (statOk)
     {
-        int perms = st.st_mode & ~S_IFMT;
-        if (perms != mode)
-        {
-            // Fix the permissions.
-            if (0 != ADUCPAL_chmod(path, mode))
-            {
-                filePermissionsChanged = true;
-                stat(path, &st);
-                Log_Error(
-                    "Failed to set '%s' file permissions (expected:%d, actual: %d)", path, mode, st.st_mode & ~S_IFMT);
-            }
-        }
-
         // Ensure that the script has the correct ownership.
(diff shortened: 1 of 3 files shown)


src/adu-shell/src/script_tasks.cpp
```diff
@@ -59,10 +59,23 @@ ADUShellTaskResult Execute(const ADUShell_LaunchArguments& launchArgs)
             {
                 filePermissionsChanged = true;
                 stat(path, &st);
-                Log_Warn(
+                Log_Error(
```
jw-msft: Should it set error value (non-zero) using taskResult.SetExitStatus and essentially "goto done" (put done: label around line 92 before return) so it does not run any more logic?

src/adu-shell/src/script_tasks.cpp
```diff
@@ -59,10 +59,23 @@ ADUShellTaskResult Execute(const ADUShell_LaunchArguments& launchArgs)
...
+        // Ensure that the script has the correct ownership.
+        struct group* grp = ADUCPAL_getgrnam(ADUC_FILE_GROUP);
+        struct passwd* p = ADUCPAL_getpwnam(ADUC_FILE_USER);
+
+        if (p != NULL && grp != NULL)
+        {
+            // Fix the ownership.
+            if (0 != ADUCPAL_chown(path, p->pw_uid, grp->gr_gid))
+            {
+                Log_Error("Failed to set '%s' file ownership to %d:%d", path, p->pw_uid, grp->gr_gid);
```
jw-msft, line 63: Should it set error value (non-zero) using taskResult.SetExitStatus and essentially "goto done" (put done: label around line 92 before return) so it does not run any more logic?

Also, for Line 76, should it be Log_Error, SetExitStatus, and also goto done?

jw-msft (approved):

src/adu-shell/src/script_tasks.cpp
Nox-MSFT: Fixed

PR #716, opened 2025-03-25
Device Update Agent v1.2 for Ubuntu Core

Nox-MSFT:
- Port Ubuntu Core Snap support from feature/snap-1 branch
- Update an Agent version to 1.2.0
- Disable diagnostic and command helper components
- Fix shelcheck errors and warnings in example-snap-update.sh


CMakeLists.txt
```diff
@@ -95,7 +95,7 @@ include (aduc_helpers)
 set_cache_with_env_or_default (ADUC_VERSION_MAJOR "1" STRING
                                "The major part of the semantic version")
 
-set_cache_with_env_or_default (ADUC_VERSION_MINOR "1" STRING
+set_cache_with_env_or_default (ADUC_VERSION_MINOR "2" STRING
```
jw-msft, line 112: should this minor ver only change in release/1.2 branch so it does not rev until merged back into develop for the commit id that would be tagged as 1.2 release?

snap/hooks/configure
```diff
@@ -0,0 +1,173 @@
...
+        fi
+
+        echo "Generating the diagnostics configuration file..."
+        if [ ! -f "$adu_conf_dir/$adu_diagnostics_conf_file" ]; then
+            echo "$sample_du_diagnostics_config" > "$adu_conf_dir/$adu_diagnostics_conf_file"
+        fi
+
+        # Create home dir
+        if [ ! -d "$adu_home_dir" ]; then
+            mkdir -p "$adu_home_dir"
```
jw-msft, line 125: was shellcheck run on all scripts?

CMakeLists.txt
```diff
@@ -212,6 +212,46 @@ set (
...
 
+set (
+    ADUC_DOWNLOAD_HANDLER_EXTENSION_DIR
+    "${ADUC_EXTENSIONS_FOLDER}/${ADUC_EXTENSIONS_SUBDIR_DOWNLOAD_HANDLERS}"
+    CACHE STRING "Path for the download handler extensions")
+
+if (ADUC_BUILD_UNIT_TESTS)
+    set (
+        EIS_DATA_PATH
+        "/tmp/eis_conn_info.txt"
```
jw-msft, line 237: release artifacts are often built with -u so this would have the wrong path in production binaries.  Instead, can have 'testdata' folder under ut folder and the copy_test_data CMake macro will be run that will copy it under a deterministic test directory (under tmp):

https://github.com/Azure/iot-hub-device-update/blob/99ab06fbdf9dc91f4015eb83fc2f4a2bf977346c/cmake/aduc_helpers.cmake#L51

jw-msft (commented):
some comments for a couple of things that should be improved

Issue #715, opened 2025-03-25
Using curl content downloader still relies on DeliveryOptimization DO-client for rootkey download


AndreRicardo-Zoetis:
We are trying to test OTA in Docker containers but noticed that even though there is an entry to use `libcurl_content_downloader`

https://github.com/Azure/iot-hub-device-update/blob/35de3285899446215a09b7b191a8401c6d03466e/scripts/docker/build_docker_image.sh#L197

The root key workflow, only seems to work with "DO", am I missing something?

https://github.com/Azure/iot-hub-device-update/blob/35de3285899446215a09b7b191a8401c6d03466e/src/rootkey_workflow/src/rootkey_workflow.c#L46-L50

Modified `/var/lib/adu/extensions/content_downloader/extension.json` to use `libcurl_content_downloader.so`

```json
{
   "fileName":"/var/lib/adu/extensions/sources/libcurl_content_downloader.so",
   "sizeInBytes":0,
   "hashes": {
        "sha256":"4M06Q8dj1JNNtcSDnftZgwPeIvHycqYJphiGsjfKpwo="
   }
}
```

In the end because `deliveryoptimization-agent` is not running on the container, the OTA fails with:

```sh
2025-03-25T13:43:14.3848Z 38423[38423] [D] Attempting download of 'http://[redacted].microsoft.com/WestUS2/rootkeypackages/rootkeypackage-1.json' using 'DO' [ADUC_RootKeyPackageUtils_DownloadPackage:114]
2025-03-25T13:43:14.3848Z 38423[38423] [I] Downloading File 'http://[redacted].microsoft.com/WestUS2/rootkeypackages/rootkeypackage-1.json' to '/var/lib/adu/downloads/a8ab014a-82ac-49f7-b405-8f5ce2c9f2d0/rootkeypackage-1.json' [DownloadRootKeyPkg_DO:22]
2025-03-25T13:43:25.3935Z 38423[38423] [E] DO error, msg: unrecognized error, code: 0x80d01001, timeout? 0 [DownloadRootKeyPkg_DO:36]
2025-03-25T13:43:25.3937Z 38423[38423] [I] Download rc: 0, erc: 0xa0101001 [DownloadRootKeyPkg_DO:51]
2025-03-25T13:43:25.3937Z 38423[38423] [E] Unable to download root key pkg. [ADUC_RootKeyPackageUtils_DownloadPackage:120]
[...]

Issue #711, opened 2025-03-21
curl content downloader does not seem to resume downloads


AndreRicardo-Zoetis:
The `curl_content_downloader.cpp` does not seem to resume downloads the same way do-client does. Any reason why?

The command built below:

https://github.com/Azure/iot-hub-device-update/blob/d226c1ef8da00daf070f001ebc7cda92725ea759/src/extensions/content_downloaders/curl_downloader/curl_content_downloader.cpp#L111-L116

Results in command:
```sh
/usr/bin/curl -o /var/lib/adu/downloads/7277db94-f0d9-4295-b41b-089709302b2e/0/my_firmware_something.bin -O http://redacted.dl.adu.microsoft.com/westus/some-instance/c2a1713966c74f91b259bd85223441db/my_firmware_something.bin
```

Would it be possible to resume downloads by passing `-C -`?

Sources:
https://curl.se/docs/manpage.html#-C
https://stackoverflow.com/a/47343997/388038

Example:
```sh
/usr/bin/curl -o /var/lib/adu/downloads/7277db94-f0d9-4295-b41b-089709302b2e/0/my_firmware_something.bin -O -C - http://redacted.dl.adu.microsoft.com/westus/some-instance/c2a1713966c74f91b259bd85223441dbmy_firmware_something.bin
```

Another smaller issue is: both `-o` and `-O` are being passed that outputs `"Warning: Got more output options than URLs"`, source: https://curl.se/mail/archive-2024-04/0017.html

From the variable `output` in `exitCode = ADUC_LaunchChildProcess("/usr/bin/curl", args, output);`
```
"  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current\n", ' ' <repeats 33 times>, "Dload  Upload   Total   Spent    Left  Speed\n\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0\r100  1003  100  1003    0     0   7670      0 --:--:-- --:--:-- --:--:--  7775\nWarning: Got more output options than URLs\n"
```

Would you accept a PR to add `-C -` ?

jw-msft:
The 2nd `-O` should not be there and yes `-C -` would be good to add.

jw-msft:
I'll take care of this as part of adding curl downloading or rootkey.

jw-msft:
@AndreRicardo-Zoetis
Draft PR: https://github.com/Azure/iot-hub-device-update/pull/717
[...]

PR #717, opened 2025-03-25
Allow use of curl to download rootkey pkg

jw-msft:
- Add ADUC_ROOTKEY_PKG_DOWNLOAD_WITH_CURL option in CMakelist.txt that sets compiler definition ROOTKEY_PKG_DOWNLOAD_USE_CURL to 1
- Add --rootkeypkg-curl option to build.sh that sets -DADUC_ROOTKEY_PKG_DOWNLOAD_WITH_CURL=true
- Add DownloadRootKeyPkg_Curl
- Conditionally compile `DownloadRootKeyPkg_Curl` or `DownloadRootKeyPkg_DO` based on ROOTKEY_PKG_DOWNLOAD_USE_CURL CompilerDefinition
- Remove `-O` (`--remote-name`) from Curl Content Downloader since -o is absolute file path to avoid `"Warning: Got more output options than URLs"`
- Add `-L` (`--location`) to curl cmd-line to handle 3xx redirects
- Add `-C -` to direct curl to auto-resume downloads
- Update tools and tests accordingly

Addresses these 2 github issues:
- https://github.com/Azure/iot-hub-device-update/issues/711
- https://github.com/Azure/iot-hub-device-update/issues/715

### DIFF
--- a/src/utils/rootkeypackage_utils/src/rootkeypackage_curl_download.cpp
+++ b/src/utils/rootkeypackage_utils/src/rootkeypackage_curl_download.cpp
@@ -49,7 +49,7 @@ ADUC_Result DownloadRootKeyPkg_Curl(const char* url, const char* targetFilePath)
             }
             joined_args += arg;
         }
-        Log_Info("Using cmdline: /usr/bin/curl '%s' ...", joined_args.c_str());
+        Log_Info("Using cmdline: /usr/bin/curl %s", joined_args.c_str());
 
         std::string output;
         int exitCode = ADUC_LaunchChildProcess("/usr/bin/curl", args, output);
--- a/src/utils/rootkeypackage_utils/src/rootkeypackage_curl_download.cpp
+++ b/src/utils/rootkeypackage_utils/src/rootkeypackage_curl_download.cpp
@@ -1,0 +1,77 @@
+/**
+ * @file rootkeypackage_curl_download.cpp
+ * @brief Implements curl download of root key packages.
+ *
+ * @copyright Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT License.
+ */
+
+#include <aduc/types/adu_core.h>
+#include <aduc/logging.h>
+#include "aduc/process_utils.hpp" // for ADUC_LaunchChildProcess
+#include <aduc/result.h>
+
+#include <string>
+#include <vector>
+
+EXTERN_C_BEGIN
+
+ADUC_Result DownloadRootKeyPkg_Curl(const char* url, const char* targetFilePath)
+{
+    ADUC_Result result = { ADUC_GeneralResult_Failure, 0 };
+
+    Log_Info("Downloading File '%s' to '%s'", url, targetFilePath);
+
+    try
+    {
+        std::vector<std::string> args;
+
+        // -L (or --location). Handle 3xx redirects. See https://curl.se/docs/manpage.html#-L
+        args.emplace_back("-L");
+
+        // -C (or --continue-at). With hyphen after, e.g. '-C -', enables auto-resuming transfers. See https://curl.se/docs/manpage.html#-C
+        args.emplace_back("-C");
+        args.emplace_back("-");
+
+        // -o (or --output). Output to file instead of stdout. See https://curl.se/docs/manpage.html#-o
+        // NOTE: -O (or --remote-name) is not needed as we already have an /absolute/ file path.
+        args.emplace_back("-o");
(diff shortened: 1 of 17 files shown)


Nox-MSFT (dismissed):

AndreRicardo-Zoetis:
Wow, this would be great!

Another interesting change would be to update `packages/CMakeLists.txt` so it could conditionally list the .deb dependencies.

Example in:
https://github.com/AndreRicardo-Zoetis/iot-hub-device-update/compare/run-in-docker...curl-rootpackage_download#diff-cb77d374cf3c80b96661c5830063aec0e2419d5680afe28d36f373787d5e7bc6

And also if it was possible in `debian/postinst` to call either curl or do-client based on `--rootkeypkg-curl`, eg 
```sh
adu_delivery_optimization_downloader_file=libdeliveryoptimization_content_downloader.so
```
or
```
adu_delivery_optimization_downloader_file=libcurl_content_downloader.so
```
Not sure if anyone is interested in mixing both curl and do-client, so to me makes sense that the option curl or do-client is applied to both root package download and content downloader.

Maybe rename `adu_delivery_optimization_downloader_file` to something like `adu_content_downloader_extension`?

https://github.com/Azure/iot-hub-device-update/blob/beda84aa16b5c5aea29d560e53b139fade3fcab2/packages/debian/postinst#L39

AndreRicardo-Zoetis:
Another question I have is, is `libcurl4-openssl-dev` really being used?

https://github.com/search?q=repo%3AAzure%2Fiot-hub-device-update%20libcurl4-openssl-dev&type=code

jw-msft:
I think at the very least `libcurl4-openssl-dev` can be replaced with `curl` (in addition to deliveryoptimization dependencies). let me make that change.

Since this reference agent is bundling BOTH curl and DO content downloaders and allows switching between them via extension registration, then the .deb pkg DEPENDS has to be the union of those.

This could in the future be changed to remove curl / DO bindings at compile-time so it's one or the other for both update payload and rootkey pkg downloading, but that would double the number of .deb packages pushed to packages.microsoft.com

jw-msft:
I've pushed commit that changes deb pkg DEPENDS from `libcurl4-openssl-dev` to `curl`

Nox-MSFT (approved):

src/utils/rootkeypackage_utils/src/rootkeypackage_curl_download.cpp
```diff
@@ -0,0 +1,77 @@
...
+            joined_args += arg;
+        }
+        Log_Info("Using cmdline: /usr/bin/curl %s", joined_args.c_str());
+
+        std::string output;
+        int exitCode = ADUC_LaunchChildProcess("/usr/bin/curl", args, output);
+        if (exitCode == 0)
+        {
+            result.ResultCode = ADUC_Result_Download_Success;
+            Log_Info("Download output:: \n%s", output.c_str());
```
AndreRicardo-Zoetis, line 59: @jw-msft I think download output should both on success or failure, otherwise is going to be hard to understand in the logs for download errors.

In the else for the failure, maybe something like this?
```c
Log_Error(
            "Curl process error, exitCode: %d\nDownload output:: \n%s",
            exitCode,
            output.c_str());
```

AndreRicardo-Zoetis:
Another comment, the equivalent of curl in do-client has a timeout of one hour.

https://github.com/Azure/iot-hub-device-update/blob/99ab06fbdf9dc91f4015eb83fc2f4a2bf977346c/src/utils/rootkeypackage_utils/src/rootkeypackage_do_download.cpp#L16

Should an equivalent option such as `--max-time` "3600" be added to the curl as well?

jw-msft:
@AndreRicardo-Zoetis see https://github.com/Azure/iot-hub-device-update/pull/736 for follow-up PR

ni-thib:
I think we are missing some test to avoid DO in non WIN32 compilation.
I had to add the attached patch, donno if it is the right way

[Don't_include_DO_while_using_CURL.patch](https://github.com/user-attachments/files/19810194/Don.t_include_DO_while_using_CURL.patch)

jw-msft:
> I think we are missing some test to avoid DO in non WIN32 compilation. I had to add the attached patch, donno if it is the right way
> 
> [Don't_include_DO_while_using_CURL.patch](https://github.com/user-attachments/files/19810194/Don.t_include_DO_while_using_CURL.patch)

Thanks for the patch. I'll see if I can incorporate it.

PR #768, opened 2025-10-16
Fix all known SDK stability issues on RPI4

jw-msft:
- Open for read to prevent EOF on client side (keep-alive descriptor) and clean it up
- Flush resp fifo and add some delay
- Remove by zlog dep by aducsdk that creeped in
- Add stderr proto traces if ADUC_SDK_DEBUG env var is set
- Add open resp fifo retries
- Increase select timeout to 500000 (500 ms)
- Retry write to resp fifo if errno == EAGAIN || errno == EWOULDBLOCK up to MAX_WRITE_RETRIES
- Use single select call to detect data ready to read and remove from loop
- Retry reads if errno == EAGAIN || errno == EWOULDBLOCK up to MAX_RETRIES with 150ms usleep between

### DIFF
--- a/src/agent/api/src/apisvc.c
+++ b/src/agent/api/src/apisvc.c
@@ -198,6 +198,7 @@ static void* aduc_apisvc_thread_proc(void* arg)
     int rdfifo = -1;
     int writefifo = -1;
     int open_read_retries = 0, open_write_retries = 0;
+    ssize_t sent = 0;
 
     ADUC_Logging_Init(ADUC_LOG_DEBUG, "aducapi");
 
@@ -244,7 +245,7 @@ static void* aduc_apisvc_thread_proc(void* arg)
 
     while (g_api_svc_thread_running)
     {
-        ssize_t n;
+        ssize_t n = 0;
         ApiWireRequestMsg msg = { 0 };
 
         open_write_retries = 0;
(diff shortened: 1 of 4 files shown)


src/agent/api/src/apisvc.c
```diff
@@ -361,12 +391,25 @@ static void* aduc_apisvc_thread_proc(void* arg)
             }
         }
 
+        if (sent > 0)
+        {
+            fsync(writefifo); // attempt to flush to the pipe before closing
+            usleep(50000); // 50 msec delay to allow client to read before closing the fifo
+        }
+
```
Nox-MSFT, line 399: Can this cause a potential race condition? (e.g., where the reader make take longer than 50000 msec to read? (or currently doing something else and not currently reading the data?)

src/utils/apiproto_utils/src/apiproto.c
```diff
@@ -214,13 +223,28 @@ ssize_t msg_send_resp(int fd, const ApiWireResponseMsg* msg)
 {
     char write_buf[RESP_MSG_READ_BUF_SIZE] = { 0 };
     ssize_t bytes_written = -1;
+    int retries = 0;
+    const int MAX_WRITE_RETRIES = 10;
```
Nox-MSFT, line 227: Should we define the same value MAX_WRITE_RETRIES & MAX_RETRIES (for recv) ?

src/utils/apiproto_utils/src/apiproto.c
```diff
@@ -232,34 +256,68 @@ ssize_t msg_send_resp(int fd, const ApiWireResponseMsg* msg)
...
         }
         if (bytes_read == 0)
         {
-            break; // EOF
+            // EOF - writer closed before we got all data
+            if (total_bytes_read == 0)
+            {
+                // No data yet, retry a few times for slow ARM systems
+                retries++;
+                usleep(150000); // 150ms
```
Nox-MSFT, line 307: Is it possible for the writer to write something, wait for 50ms, then close the pipe? before this 150ms sleep is over?

Nox-MSFT (approved):
Approved, w/ some questions for clarification.

PR #806, opened 2026-01-08
Pass workflowData to PrepareScriptArguments

AndreRicardo-Zoetis:
Pass workflowData to PrepareScriptArguments to make it possible to pass custom data from workflowData to the scripts.

This PR looks like has no impact in ADU overall but this change would really help us in our customization for Zoetis as we need to pass extra arguments from the script handler.

Thanks to @HenrikAndreasen-Zoetis


AndreRicardo-Zoetis:
Any feedback on this change? Thank you.

chgennar (approved):
Thank you for the contribution, Andre!

PR #813, opened 2026-02-20
feat: add run_coverage.sh script to generate code coverage report

whereiseva:


scripts/run_coverage.sh
```diff
@@ -0,0 +1,171 @@
+#!/usr/bin/env bash
```
nonsocode, line 1: Is it necessary for this script to build the project? It seems like the `build.sh` script's job is primarily building. Wouldn't this script be best suited for coverage alone?

scripts/build.sh
nonsocode, line 404: nit: I think coverage parameters can be parsed within this argument processing section. The coverage args can also determine the build params.

Coverage arguments could be passed into the build script with the `--coverage` prefix. e.g  --report-only could be --coverage-report-only

whereiseva: we can pass more parameters to run_converage.sh script through build script.

if we run `build.sh --coverage --report-only`, it will be `run_coverage.sh --report-only`.

scripts/run_coverage.sh
whereiseva: I keep build step in this script because: 
- overage is only trustworthy when binaries are built with instrumentation flags; if build is external, it’s easy to run tests on stale or non-instrumented artifacts and get misleading results.
- having build step in the script, we can easily add instrumentation flags without changing build script much. 

Instrumented build settings:
export CFLAGS="--coverage -O0 -g"
export CXXFLAGS="--coverage -O0 -g"
export LDFLAGS="--coverage"
build.sh -u -t Debug
 
The script does full end-to-end by default, and supports report-only mode if we're sure we already built/tested the latest with instrumentation.

PR #829, opened 2026-02-26
Adding unit tests for extensions

mfsmoot:
This PR improves the unit test coverage of /src/extensions. It brings the coverage percentage from 19.90% to 62.45%. 

Notes:
- The component_enumerators file is an example and will be moved into the /docs folder
- download_handler_plugin.cpp can't reach 85% coverage because there is some dead code that can't be reached.

Remaining low-coverage lines are mostly:
- exception-only branches requiring forced throws from static/library calls,
- dlopen/dlsym error permutations,
- environment/system-user-dependent flows (adu-shell, privileged FS/runtime behavior),
- hard-to-trigger internal failure branches without seams/stubs.

This is near the limit of what we can achieve with non-mock unit tests. Pushing further without mocks would likely become flaky or host-dependent rather than reliable unit coverage.

image


mfsmoot:
@microsoft-github-policy-service agree [company="Microsoft"]

mfsmoot:
@microsoft-github-policy-service agree company="Microsoft"

whereiseva (dismissed):

whereiseva (dismissed):

whereiseva (approved):

PR #830, opened 2026-02-26
adu_workflow unit test improvements

hmmorales:
this unit test update brings up the coverage to 75%
image


src/adu_workflow/src/agent_workflow.c
```diff
@@ -1347,6 +1384,12 @@ void ADUC_Workflow_MethodCall_Idle(ADUC_WorkflowData* workflowData)
  */
 ADUC_Result ADUC_Workflow_MethodCall_ProcessDeployment(ADUC_MethodCall_Data* methodCallData)
 {
+    if (methodCallData == NULL || methodCallData->WorkflowData == NULL)
```
Nox-MSFT: @hmmorales , please ensure that we're setting meaningful Extended Result Code when returning failed results.

src/adu_workflow/src/agent_workflow.c
```diff
@@ -1401,18 +1450,27 @@ ADUC_Result ADUC_Workflow_MethodCall_Download(ADUC_MethodCall_Data* methodCallDa
...
+        {
+            goto done;
+        }
     }
 
     Log_Info("Using sandbox %s", workFolder != NULL ? workFolder : "(null)");
 
     ADUC_Workflow_SetUpdateState(workflowData, ADUCITF_State_DownloadStarted);
 
+    if (updateActionCallbacks->DownloadCallback == NULL)
```
Nox-MSFT: Several of these null checks may be unnecessary since the caller is expected to set these values. It might be cleaner to consolidate validation at the beginning of the function.

src/adu_workflow/src/agent_workflow.c
```diff
@@ -1290,6 +1316,11 @@ void ADUC_Workflow_SetInstalledUpdateIdAndGoToIdle(ADUC_WorkflowData* workflowDa
  */
 void ADUC_Workflow_MethodCall_Idle(ADUC_WorkflowData* workflowData)
 {
+    if (workflowData == NULL)
```
Nox-MSFT: @hmmorales , These values are expected to be set by the caller. If we still check for null here, I’d recommend also logging an error to improve observability.

mfsmoot (approved):

PR #881, opened 2026-06-01
install-deps: pin delta build to newer GCC on Ubuntu 20.04 arm64

Nox-MSFT:
The delta library's src/native/build.sh hard-codes triplet-prefixed compiler paths (e.g. /usr/bin/aarch64-linux-gnu-g++) and passes them to CMake via -DCMAKE_C_COMPILER / -DCMAKE_CXX_COMPILER, ignoring the CC/CXX env vars and the unprefixed update-alternatives we already set.

On Ubuntu 20.04 arm64 the default aarch64-linux-gnu-g++ is GCC 9, which lacks the C++20 <span> header that the delta sources require, so the build fails with: fatal error: span: No such file or directory.

Fix: patch the cloned build.sh to honor CC/CXX as fallback defaults, and set CC/CXX (scoped to the child process) only when the triplet's default g++ is actually older than the gcc_ver we just installed. This leaves debian11-arm64 (g++ 10), ubuntu2204-arm64 (g++ 11) and amd64 (already aliased via alternatives) untouched.

### DIFF
--- a/scripts/install-deps.sh
+++ b/scripts/install-deps.sh
@@ -782,6 +782,23 @@ do_install_delta() {
         sed -i 's/target_link_libraries(dumpextfs PRIVATE ${E2FSPROGS_LIBRARIES})/target_link_libraries(dumpextfs PRIVATE ${E2FSPROGS_LIBRARIES} ${COM_ERR_LIBRARIES})/' "$dumpextfs_cmake"
     fi
 
+    # Patch the delta build.sh so it honors CC/CXX env overrides as fallback
+    # defaults instead of always passing the hard-coded triplet compiler paths
+    # to CMake. This lets us pin a newer GCC (below) on distros whose
+    # triplet-prefixed compiler is too old for C++20 (e.g. Ubuntu 20.04 arm64
+    # ships /usr/bin/aarch64-linux-gnu-g++ at GCC 9, which lacks <span>).
+    # When CC/CXX are unset, the ${CC:-...} fallback evaluates to the original
+    # hard-coded path, so this patch is a no-op on healthy distros.
+    local delta_build_sh="$delta_dir/src/native/build.sh"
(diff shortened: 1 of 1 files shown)


chgennar (approved):

Nox-MSFT:
### Rebased onto latest `develop`

Force-pushed a rebase of this branch onto current `develop` (was ~10 commits behind). The branch now also contains:

- **#885** ("retry apt steps to absorb transient mirror failures"), and
- the **`ubuntu:20.04` + `arm64` CI exclusion** added on `develop`.

#### About the earlier `container-job (debian:11, arm64)` failure

That failure was **not** a code/compiler problem — it was a transient apt mirror error while fetching `gdb`:

```
Err:9 ... gdb arm64 10.1-1.7  Error reading from server - read (104: Connection reset by peer)
E: Failed to fetch .../gdb_10.1-1.7_arm64.deb
```

`debian:11` ships **g++ 10**, which already supports the C++20 `<span>` this PR is about, so debian11-arm64 is a healthy target. The job failed only because this branch predated **#885**'s apt-retry hardening. The rebase pulls that in, so the re-triggered run should be resilient to the flaky download. (No need to drop `debian:11, arm64` from the matrix.)

#### Note on scope — merge vs. supersede

`develop`'s CI now **excludes `ubuntu:20.04` + `arm64`** (the exact combination this PR fixes), citing the GCC‑9/`<span>` issue and Ubuntu 20.04 being EOL (April 2025). As a result, after the rebase this PR has **no effect on CI**; it only benefits **manual/local delta-handler builds on Ubuntu 20.04 arm64**.

Two coherent options for maintainers:
- **Merge** — to keep manual Ubuntu 20.04 arm64 delta builds working, or
- **Close as superseded** — if Ubuntu 20.04 arm64 is no longer a supported target given the EOL exclusion.

